PR #3, opened 2015-10-12
Shared secret auth token

parente:
* [X] Traitlet for specifying the secret on app start (command line or config file)
* [X] Test case checking for 401 on all kernel resource requests when enabled
* [x] Options doc'ed in the README somehow (or ref --help-all)
* [x] Env var support for it to ease use from Docker (minrk's approach: https://github.com/jupyter/oauthenticator/blob/4db3cbb7d7fb5878976946c69944dba6ba0f968e/oauthenticator/oauth2.py#L83)

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -1,5 +1,6 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+import os
 
 from traitlets import (
     Unicode, Integer
@@ -31,18 +32,26 @@ class KernelGatewayApp(JupyterApp):
         Provisions kernels and bridges Websocket communication
         to/from them.
     '''
-
-    port = Integer(8888, config=True,
-        help="Port on which to listen"
+    port_env = 'KG_PORT'
+    port = Integer(config=True,
+        help="Port on which to listen (KG_PORT env var)"
     )
+    def _port_default(self):
+        return int(os.getenv(self.port_env, 8888))
 
-    ip = Unicode('0.0.0.0', config=True,
-        help="IP address on which to listen"
+    ip_env = 'KG_IP'
+    ip = Unicode(config=True,
+        help="IP address on which to listen (KG_IP env var)"
     )
+    def _ip_default(self):
+        return os.getenv(self.ip_env, '0.0.0.0')
 
-    auth_token = Unicode('', config=True,
-        help='Authorization token required for all requests'
+    auth_token_env = 'KG_AUTH_TOKEN'
+    auth_token = Unicode(config=True,
+        help='Authorization token required for all requests (KB_AUTH_TOKEN env var)'
     )
+    def _auth_token_default(self):
+        return os.getenv(self.auth_token_env, '')
 
     def initialize(self, argv=None):
         '''
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -2,12 +2,9 @@
 # Distributed under the terms of the Modified BSD License.
 import os
 
-from traitlets import (
-    Unicode, Integer
-)
+from traitlets import Unicode, Integer
 
 from jupyter_core.application import JupyterApp
-# from jupyter_core.paths import jupyter_runtime_dir
 from jupyter_client.kernelspec import KernelSpecManager
 
 # Install the pyzmq ioloop. This has to be done before anything else from
@@ -18,10 +15,8 @@ ioloop.install()
 import tornado
 from tornado import httpserver
 from tornado import web
-# from tornado.log import LogFormatter, app_log, access_log, gen_log
 
 from .services.kernels.handlers import default_handlers as default_kernel_handlers
-# import notebook.services.kernels.handlers as kernel_handlers
 from notebook.services.kernels.kernelmanager import MappingKernelManager
 
 class KernelGatewayApp(JupyterApp):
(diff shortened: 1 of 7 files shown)


rgbkrk:
Oh hey, slick!

parente:
@rgbkrk care to review before merge? @bpburns?

kernel_gateway/gatewayapp.py
```diff
@@ -19,7 +20,8 @@
 from tornado import web
 # from tornado.log import LogFormatter, app_log, access_log, gen_log
 
-import notebook.services.kernels.handlers as kernel_handlers
+from .services.kernels.handlers import default_handlers as default_kernel_handlers
+# import notebook.services.kernels.handlers as kernel_handlers
```
rgbkrk: You can delete the commented import

kernel_gateway/gatewayapp.py
```diff
@@ -30,16 +32,32 @@ class KernelGatewayApp(JupyterApp):
...
+        return int(os.getenv(self.port_env, 8888))
 
-    port = Integer(8888, config=True,
-        help="The port the notebook server will listen on."
+    ip_env = 'KG_IP'
+    ip = Unicode(config=True,
+        help="IP address on which to listen (KG_IP env var)"
     )
+    def _ip_default(self):
+        return os.getenv(self.ip_env, '0.0.0.0')
```
rgbkrk: I'm never a fan of services that listen on all interfaces by default, but that's also the easiest to work with in the docker case.

parente: No. This is a good point. I'll default it to 127.0.0.1.

kernel_gateway/services/kernels/handlers.py
```diff
@@ -0,0 +1,26 @@
...
+    def prepare(self):
+        '''
+        If kg_auth_token is present in the application settings,
+        raises 401 if the request Authorization header does not contain
+        the value "Token <kg_auth_token>".
+        '''
+        server_token = self.settings.get('kg_auth_token')
+        if server_token:
+            client_token = self.request.headers.get('Authorization')
+            if client_token != 'Token %s' % server_token:
```
rgbkrk: I'm not sure what the standard practice, but from working with the GitHub API I've gotten used to `Token` being lowercase `token`:

```
curl -H "Authorization: token 6c0b8049fa7de55e31becc84dc970498" https://api.github.com
```

(that's a randomly generated token by the way)

kernel_gateway/services/kernels/handlers.py
```diff
@@ -0,0 +1,26 @@
...
+    '''
+    Adds token authentication to tornado.web.RequestHandlers.
+    '''
+    def prepare(self):
+        '''
+        If kg_auth_token is present in the application settings,
+        raises 401 if the request Authorization header does not contain
+        the value "Token <kg_auth_token>".
+        '''
+        server_token = self.settings.get('kg_auth_token')
```
rgbkrk: Not that you have to, but I tend to decorate a function with `@property` for settings like this.

```
@property
def kg_auth_token():
  return self.settings.get('kg_auth_token')
```

Then I just use it as 

```
server_token = self.kg_auth_token
```

I'd also prefer to not have the `kg_` prefix on the code itself if it already is part of the kernel gateway. That's a nit though, so feel free to ignore.

rgbkrk:
:scream: This is great, I'm *so* looking forward to using this. :smiley:

kernel_gateway/services/kernels/handlers.py
parente: Switching to lowercase.

kernel_gateway/services/kernels/handlers.py
parente: I prefixed the settings values to avoid conflicts with Tornado settings. I don't plan to propagate them further.

For the moment, I won't add the properties because I know for the CORS settings (next PR) I'm already going to have to refactor to handle the Websocket handler differently than the HTTP ones. I'll see what makes sense then.

kernel_gateway/tests/test_gatewayapp.py
```diff
@@ -56,7 +71,69 @@ def test_headless(self):
...
+        # Request websocket connection without the token
+        ws_url = 'ws://localhost:{}/api/kernels/{}/channels'.format(
+            self.get_http_port(),
+            url_escape(kernel['id'])
+        )
+        # No option to ignore errors so try/except
+        try:
+            ws = yield websocket_connect(ws_url)
+        except Exception as ex:
+            self.assertEqual(ex.code, 401)
```
rgbkrk: Should this test to make sure that the `Exception` does get raised? (it won't test it if no `Exception` occurred)

rgbkrk:
Though it's not used in the Jupyter code base, [there is a way to `fetch` directly without using `http_client.fetch` + `get_url`](http://www.tornadoweb.org/en/latest/testing.html#tornado.testing.AsyncHTTPTestCase). The code is already subclassing `AsyncHTTPTestCase` and providing `get_app`, so the boilerplate pieces are already there.

For instance, instead of this code...

```
yield self.http_client.fetch(self.get_url('/api/kernels'))
self.assertEqual(response.code, 200)
```

... you can write:

```
response = self.fetch('/api/kernels')
self.assertEqual(response.code, 200)
```

parente:
Unfortunately, the shortcut produced a `RuntimeError: IOLoop is already running` error when I tried it. I think it's something to do with the zmqloop being used for the kernel manager but instead of digging in to fully understand, I punted on the shortcut and went with the longhand that worked.

kernel_gateway/tests/test_gatewayapp.py
parente: Added assertion in an else.

rgbkrk:
Oh ok, that may have been the decision in the notebook too. I've used tornado in other contexts, so I had to wonder.

parente:
Think I've addressed all the comments here. If anything comes up, can continue to improve on future pulls.

rgbkrk:
Awesome, thanks for pinging me!

PR #7, opened 2015-10-13
Add /api and /api/kernelspecs

parente:
The /api source is important for knowing the API version. The /api/kernelspecs is important for knowing what kernels are available (and maybe one day fetching kernel frontend assets).

### DIFF
--- a/kernel_gateway/base/handlers.py
+++ b/kernel_gateway/base/handlers.py
@@ -1,0 +1,14 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import notebook.base.handlers as notebook_handlers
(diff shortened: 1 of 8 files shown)


kernel_gateway/base/handlers.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import notebook.base.handlers as notebook_handlers
+from ..mixins import TokenAuthorizationMixin, CORSMixin
+
+class APIVersionHandler(TokenAuthorizationMixin,
```
parente: Not sure if this API metadata should or should not protected by auth too. For now, erring on the side of safety: protect it.

PR #15, opened 2015-10-14
Enable travis

parente:
* Add make install target
* Only expose docker ports on dev, bash make targets

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,10 @@
-sudo: required
-
 language: python
-
 python:
+    - 2.7
+    - 3.3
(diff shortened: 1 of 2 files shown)


parente:
I went with non-containerized testing on Travis due to issues with Docker caps set on their infrastructure and then problems getting the tornado test runner to work. I originally traced the errors to tini so the details are over here https://github.com/krallin/tini/issues/9#issuecomment-148078280

At any rate, all the tests are passing now on Travis outside the container. And we get to test against multiple version of Python, not just the one in the container.

PR #18, opened 2015-11-06
Option to seed kernel memory with notebook content

### LINKED ISSUE #10: Option to pre-seed kernel memory
Given a URI of a notebook, execute its contents on launched kernels, either immediately or when the first websocket connection is established.

parente:
Fixes #10


parente:
Updated PR with the requests requirement.

parente:
@jtyberg, @rgbkrk want to take a look?

One note: The new kernel manager in this PR needs to execute an entire notebook's worth of code on the kernel it launches. While doing so it has to block the response, otherwise the requesting client can start accessing the notebook before all code has run. (TIL: commands are not sequenced across client connections to kernels, only within a given connection.) This is problematic because the `start_kernel` code path is synchronous in the notebook project which means that while the kernel manager is blocking waiting for all cells to execute, the entire kernel gateway server is also blocked. This if fine if we're pairing 1 gateway with 1 kernel and scaling a container level, but not if otherwise.

This has come up before, too, when we've experimented with remote kernels. We don't want the whole notebook server to block while a remote kernel is launching.

I wonder if it's worth a PR against notebook to see if the `start_kernel` code path could be turned into a coroutine. I don't know how much leakage that causes of tornado concepts into subclasses of jupyter_client or if that even matters. @minrk?

rgbkrk:
This is an interesting one for dashboards I presume?

minrk:
We use tornado coroutines/Futures to do some async stuff in the existing notebook server. You can use that to make this non-blocking.

parente:
@minrk The problem here is that we're calling `start_kernel` on a subclass of `MappingKernelManager` from the notebook which is not decorated as a coroutine. I think that means that even if our `start_kernel` method in the subclass is made non-blocking, we're still going to be blocked as soon as we call into the base class (and the base class calls subprocess.Popen somewhere down the line.)

Correct me if I'm wrong, but wouldn't the change need to be made extensively throughout the notebook code?

minrk:
I think we'll probably want to allow several things to become async in the notebook handlers that we don't currently. It's typically pretty simple to wrap things in `gen.maybe_future`, but you are right that it does need to propagate up the call stack to the top-level Handler to expect coroutines. It shouldn't need to be a big change, though.

minrk:
jupyter/notebook#711 should allow these operations to be async. You can make them sync in the meantime, or apply a similar patch to the handlers here.

parente:
Nice. Thanks @minrk for adding that so quickly. When it lands, we'll take advantage of it here.

/cc @danielballan might be interested too for remotekernel

danielballan:
Thanks, @parente. Yes, I think this could help with the performance of conda-kernels as well -- currently, there is a noticeable lag while it calls out to `conda env list`.

jtyberg:
I kicked the tires on this a bit, using a simple notebook that imports pandas and the `jupyter/datascience-notebook` stack.

```
DOCKER_ARGS="-e KG_SEED_NOTEBOOK=https://gist.githubusercontent.com/jtyberg/7c5c6ea7b541a821c298/raw/b0332b43d7058de6522bf4b4ec9a33d50151e805/pandas_yahoo.ipynb" \
IMAGE=jupyter/datascience-notebook \
make dev
```

Creating a kernel fetches and runs the notebook

```
curl -X POST http://192.168.99.101:8888/api/kernels
```

Everything seems to work fine, as long as the seed notebook can actually run (dependencies are installed in whatever container you're running, etc.).  I'm not clear on what the UX should be if a seed notebook cannot run.  Currently it responds with an error, but creates a kernel anyway.  I guess we can sort that out later?

kernel_gateway/services/kernels/manager.py
```diff
@@ -0,0 +1,28 @@
...
+        # before returning the kernel ID to the requesting client
+        if seed_source is not None and kernel_id:
+            # Get a client for the new kernel
+            kernel = self.get_kernel(kernel_id)
+            client = kernel.client()
+            for code in seed_source:
+                # Execute every code cell and wait for each to succeed or fail
+                client.execute(code)
+                msg = client.shell_channel.get_msg(block=True)
+                if msg['content']['status'] != 'ok':
```
parente: @jtyberg I thought this line would handle errors in execution. Perhaps not. And it definitely doesn't go and shut the kernel back down yet, so it leaks.

I can look at better cleanup before we merge.

parente: Code above does catch errors. Added a test to confirm it. Also added logic to ensure notebook language matches available kernel language.

parente:
@jtyberg Addressed your comments

1. If kernel cannot run the seed notebook completely, shut down the kernel.
2. If requested kernel language does not match that of the seed notebook, don't try to exec the notebook on the kernel.
3. If a kernel spec for the seed notebook is not installed, error on startup (no way to use the seed).

Tests for everything. Fixed Python 2 failing tests as well.

jtyberg:
Updates look good.  Thanks @parente.

PR #31, opened 2015-11-19
Example kernel + kernel gateway docker image

poplav:
New docker stack for a headless kernel. It uses the new kernel gateway to provide websocket access without exposing the whole notebook UI.


docker-stacks/all-spark-notebook/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) Jupyter Development Team.
+FROM jupyter/all-spark-notebook:dbb9c7b50531
+
+MAINTAINER Jupyter Project <jupyter@googlegroups.com>
+
+# Install Kernel Gateway
+RUN pip install jupyter_kernel_gateway==0.1.0
```
parente: Because you're inheriting from jupyter/all-spark-notebook, you should only need to include the RUN, the MAINTAINER, and this RUN. But you need to switch to `USER jovyan` before the RUN so that it installs with proper permissions set for the non-privileged user in the container image and set it back to `USER root` after.

Everything else below this line (and all the associated files) shouldn't be necessary. You can remove the README too since this is meant as a demo at the moment more than anything.

parente:
Thanks for the update @poplav. That's better. Dawns on me though that the image is still going to run the `start-notebook.sh` script from the base image whereas now we want it to run `jupyter kernelgateway` and pass through any command line args passed to the docker run command to it. It would be good to make that the CMD in the dockerfile (or a script that does the passthrough). It doesn't need all the user switching and permission changes that are in the original `start-notebook.sh`. But it should run as `jovyan` not as root. Not sure if keeping the last USER statement as jovyan is right or not given that the ENTRYPOINT in the base image is `tini` and if that needs to run as root or not. (Try it).

Also, could you please move the Dockerfile under the path `etc/docker-stacks/all-spark-kernels`. Thanks.

@Lull3rSkat3r can help you out with the details since he's now a kernel gateway :100: expert :100:  from working on issue #32.

etc/docker-stacks/all-spark-notebook/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+MAINTAINER Jupyter Project <jupyter@googlegroups.com>
+
+USER jovyan
+
+# Install Kernel Gateway
+RUN pip install jupyter_kernel_gateway==0.1.0
+
+# Configure container startup
+ENTRYPOINT ["jupyter", "kernelgateway"]
```
parente: The base image uses tini as the entrypoint which we need to keep because tini makes sure that orphaned child processes are adopted properly whereas jupyter does not. 

By default, I think we can do:

```
ENTRYPOINT ["tini", "--", "jupyter", "kernelgateway"]
CMD ["--KernelGatwayApp.ip", "0.0.0.0"]
```

But then users can override it on run like:
```
docker run -d -P jupyter-incubator/kernel_gateway --KernelGatewayApp.seed_notebook=foo.ipynb # etc.
```

parente:
Thanks for the updates. This is good enough to go in. I'll probably rename it `all-spark-kernels` once it goes in since it's no longer a notebook per se.

Issue #17, opened 2015-10-21
Option(s) to control channel access


parente:
Option to limit access to one or more Jupyter protocol channels: shell, iopub, control. Maybe to limit to specific comm channel names within shell too.

Use case: Restricting kernel access to listen-only.

rgbkrk:
Listen only is a good case for iopub and using SSE instead of websockets.

parente:
I think supporting more than websocket alone is going to be one of the more interesting aspects of the gateway. I chatted with some folks at Jupyter Day about the idea of combining the idea in #10 with, say, the ability to front specific notebook content (functions? cells?) with a pure HTTP API. Kernels then become little microservice engines.

At this point, I don't mind lumping a lot of different gateway features in the ref impl in this repo. Part of "graduation" will be figuring out if things should be refactored into separate projects.

PR #32, opened 2015-11-23
[WIP] Issue 29: HTTP API Support

Lull3rSkat3r:
(c) Copyright IBM Corp. 2015

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -262,7 +262,7 @@ class KernelGatewayApp(JupyterApp):
             kernel_id = self.kernel_manager.start_kernel()
             kernel_client = self.kernel_manager.get_kernel(kernel_id).client()
             kernel_client.start_channels()
-            # Get the kernel info message
+            kernel_client.wait_for_ready()
 
             endpoints = self._load_api_notebook(self.api)
             for uri in endpoints:
@@ -272,13 +272,6 @@ class KernelGatewayApp(JupyterApp):
 
             indicator = re.compile('#\s+([A-Z]+)\s+(\/.*)+')
             notebook = nbformat.read(self.api, 4)
-
-            started = False
-            while not started:
-                # TODO: Investigate timining issue where kernel never starts up to meet the exit this loop
-                iopub_message = kernel_client.get_iopub_msg(block=True)
-                if iopub_message['msg_type'] == 'status' and iopub_message['content']['execution_state'] == 'idle':
-                    started = True
 
             for cell in notebook.cells:
                 # is its source annotated?
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -3,6 +3,7 @@
 import os
 import re
 import nbformat
+import sys
 
 try:
     from urlparse import urlparse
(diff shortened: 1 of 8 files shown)


kernel_gateway/gatewayapp.py
```diff
@@ -144,9 +146,23 @@ def _default_kernel_name_default(self):
...
+    )
+    def _list_kernels_default(self):
+        return os.getenv(self.list_kernels_env, 'False') == 'True'
+
+    api_env = 'KG_API'
+    api = Unicode(config=True,
+        help='Designates that the kernel gateway will serve APIs from this notebook (KG_API env var)'
+    )
+    def _api_default(self):
+        return os.getenv(self.api_env)
```
parente: Probably want this returning a default string representing the "notebook kernel API + jupyter protocol over websocket" option. Why: we've got one protocol+api, you guys are working on a second (notebook defined API over HTTP), and there might be more coming down the pipe (e.g., see #17). Best to plan for extensibility here.

Make the help print the available valid options. You can get it to show the default too: see the others where I set the value + used the default method.

kernel_gateway/services/notebooks/handlers.py
```diff
@@ -0,0 +1,85 @@
...
+from .request_utils import *
+
+class NotebookAPIHandler(tornado.web.RequestHandler):
+    get_source = None
+    put_source = None
+    post_source = None
+    delete_source = None
+    kernel_client = None
+    execution_timeout = 5
+    def initialize(self, sources, kernel_client):
```
parente: IIRC `initialize` is invoked per instance of this class in tornado, and *every web request spawns its own instance*. That's why the handler map is to classes, not instances.

Definitely don't want to be doing this parsing over and over again. Best to keep these out of the class at the module level (singleton) or as class vars.

kernel_gateway/gatewayapp.py
```diff
@@ -213,17 +258,47 @@ def init_webapp(self):
...
-            default_kernelspec_handlers +
-            default_base_handlers
-        ):
-            # Create a new handler pattern rooted at the base_url
-            pattern = url_path_join(self.base_url, handler[0])
-            # Some handlers take args, so retain those in addition to the
-            # handler class ref
-            new_handler = tuple([pattern] + list(handler[1:]))
-            handlers.append(new_handler)
+        if self.api:
```
parente: I know it's early and you haven't refactored yet, but I think the logic below should ultimately wind up in a function or functions defined in the `notebooks/handlers` module to encapsulate the parsing logic + execution logic in one spot? Invoke the function from here, but define it there.

Compelling reason to do it: I think these "handlers" are going to become part of a kind of lite plugin system. As long as you write a Python package/module with such-and-such API, you can plug it into kernel gateway to define new ways of talking to kernels.

nitind: We're definitely going to do some refactoring--the kernel spec resides in the notebook, but we're currently set to only read that after the KernelManager is already created.

kernel_gateway/services/notebooks/handlers.py
nitind: At the moment, the init_webapp function of gatewayapp.py is sending the parsed cell source and URI to Tornado when registering the hander class, so however messy it is right now, we're only parsing things once.

PR #33, opened 2015-11-23
Enable python2.7 testing in dev container

parente:
Re-enable travis to see if things work there too


parente:
@Lull3rSkat3r and @nitind both had a look and said :shipit:

Issue #29, opened 2015-11-18
Options to enable HTTP access to kernels


parente:
I've got it in my head that the kernel gateway can become a vehicle for providing many different ways of accessing kernels, both in terms of API and protocol. Right now, it supports HTTP to the notebook kernel CRUD API and Websockets carrying Jupyter protocol messages. We've talked about limiting channel access and doing HTTP to the notebook kernel CRUD API and SSE carrying Jupyter protocol iopub messages only.

I've got this (semi-crazy) idea that we could also optionally:

0. Turn off the websocket / kernel CRUD API
1. Have the kernel gateway read cell annotations from a notebook file on startup (option already supported)
2. Use those annotations to define HTTP endpoints referring to code cell to execute when they're invoked (new)
3. When one of those endpoints receives an HTTP request, have the kernel gateway send the request body/query args to the kernel (e.g., by executing them in an assignment statement, ...)
4. After sending the args, have the kernel gateway execute the cell code associated with the HTTP endpoint on the kernel and read its output (e.g., iopub stream message, ...)
5. Respond to the HTTP request with the output and/or error status appropriately

Pack the notebook + kernel gateway up in a container, scale that container image, and you've got a to use notebooks as a place to quickly prototype microservices that run on kernels.

parente:
Whole bunch of brainstorming with @nitind and @Lull3rSkat3r on this today. Dumping the notes here. Nothing set in stone and we're totally open to getting feedback from others. (I think it'll be easier to discuss with the community when we have something working though to show the concept.)

### What's the tenancy / scalability of kernel microservices?

A kernel is single threaded and handles one request at a time. To scale, you scale the notebook+kernel gateway+kernel pattern at a container level.

-----------------------------

### What if the kernel is stateful across requests?

The burden of ensuring the API is stateless falls on the notebook author.

-----------------------------

### What's the start command?

`jupyter kernelgateway --KernelGatewayApp.api=kernel-http --KernelGatewayApp.ip=0.0.0.0 --KernelGatewayApp.seed_uri=my_annotated_notebook.ipynb --KernelGatewayApp.max_kernels=1 --KernelGateway.prespawn_count=1`

"Run the kernel gateway and immediately launch one kernel. Run the (unannotated cells?) contents of my_annotated_notebook.ipynb on that kernel when it launches. Expose a HTTP interface to the annotated cells in the my_annotated_notebook notebook."

Of course, we can simplify this over time.

-----------------------------

[...]

PR #35, opened 2015-11-24
[WIP] [Issue 29]: Options to enable HTTP access to kernels

nitind:
Refactored with changes from @Lull3rSkat3r and @parente:
   * Use kernel client library to wait for the kernel to start,
       draining shell output when executing non-API cells
   * Added path params for access from API functions.
   * Fixed encoding issue with args parsing.
   * Added scotch notebook example.
   * Add parsing of query parameters
   * Can access request body from cells
   * Body assignment and API cell indicator respect notebook kernel spec

Ref issue #29

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -232,10 +232,10 @@ class KernelGatewayApp(JupyterApp):
         '''
         # Redefine handlers off the base_url path
         handlers = []
-
+        self.kernel_client = None
         if self.api == 'notebook-http':
             kernel_id = self.kernel_manager.start_kernel()
-            kernel_client = self.kernel_manager.get_kernel(kernel_id).client()
+            self.kernel_client = self.kernel_manager.get_kernel(kernel_id).client()
             # discover the notebook endpoints and their implementations
             endpoints = self.kernel_manager.endpoints()
             sorted_endpoints = self.kernel_manager.sorted_endpoints()
@@ -244,7 +244,7 @@ class KernelGatewayApp(JupyterApp):
                 parameterized_path = parameterize_path(uri)
                 parameterized_path = url_path_join(self.base_url, parameterized_path)
                 self.log.info('Registering uri: {}, methods: ({})'.format(parameterized_path, list(endpoints[uri].keys())))
-                handlers.append((parameterized_path, NotebookAPIHandler, {'sources' : endpoints[uri], 'kernel_client' : kernel_client, 'kernel_name' : self.kernel_manager.seed_kernelspec}))
+                handlers.append((parameterized_path, NotebookAPIHandler, {'sources' : endpoints[uri], 'kernel_client' : self.kernel_client, 'kernel_name' : self.kernel_manager.seed_kernelspec}))
         elif self.api == 'jupyter-websocket':
             # append tuples for the standard kernel gateway endpoints
             for handler in (
@@ -314,5 +314,8 @@ class KernelGatewayApp(JupyterApp):
             kids = self.kernel_manager.list_kernel_ids()
             for kid in kids:
                 self.kernel_manager.shutdown_kernel(kid, now=True)
+
+        if self.kernel_client:
+            self.kernel_client.stop_channels()
 
 launch_instance = KernelGatewayApp.launch_instance
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -259,7 +259,7 @@ class KernelGatewayApp(JupyterApp):
                 new_handler = tuple([pattern] + list(handler[1:]))
                 handlers.append(new_handler)
         else:
-            sys.exit('Invalid API value, valid values are jupyter-websocket and notebook-http')
+            raise ValueError('Invalid API value, valid values are jupyter-websocket and notebook-http')
 
         self.web_app = web.Application(
             handlers=handlers,
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -163,6 +163,10 @@ class KernelGatewayApp(JupyterApp):
     def _api_default(self):
         return os.getenv(self.api_env, 'jupyter-websocket')
 
+    def _api_changed(self, name, old, new):
+        if new not in ['notebook-http', 'jupyter-websocket']:
+            raise ValueError('Invalid API value, valid values are jupyter-websocket and notebook-http')
+
     def _load_notebook(self, uri):
         '''
         Loads a local or remote notebook. Raises RuntimeError if no installed
@@ -258,8 +262,6 @@ class KernelGatewayApp(JupyterApp):
                 # handler class ref
                 new_handler = tuple([pattern] + list(handler[1:]))
                 handlers.append(new_handler)
-        else:
-            raise ValueError('Invalid API value, valid values are jupyter-websocket and notebook-http')
 
         self.web_app = web.Application(
             handlers=handlers,
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -309,4 +309,10 @@ class KernelGatewayApp(JupyterApp):
             self.io_loop.stop()
         self.io_loop.add_callback(_stop)
 
+    def shutdown(self):
+        if self.kernel_manager:
+            kids = self.kernel_manager.list_kernel_ids()
+            for kid in kids:
+                self.kernel_manager.shutdown_kernel(kid, now=True)
+
 launch_instance = KernelGatewayApp.launch_instance
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -1,7 +1,9 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+
 import os
 import nbformat
+import sys
 
 try:
     from urlparse import urlparse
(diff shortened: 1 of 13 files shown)


parente:
Does this replace PR #32?

nitind:
Yes.

kernel_gateway/gatewayapp.py
```diff
@@ -152,9 +153,16 @@ def _default_kernel_name_default(self):
     def _list_kernels_default(self):
         return os.getenv(self.list_kernels_env, 'False') == 'True'
 
+    api_env = 'KG_API'
+    api = Unicode(config=True,
+        help='Designates that the kernel gateway will serve APIs from this notebook (KG_API env var)'
```
parente: Is there a choices traitlet? If so, this should use it. Either way, the currently supported options should be "jupyter-websocket" and "notebook-http". Anything else should be an error. The help string needs to reflect the available options if you can't define a set of choices in a traitlet.

kernel_gateway/gatewayapp.py
```diff
@@ -152,9 +153,16 @@ def _default_kernel_name_default(self):
...
+    )
+    def _api_default(self):
+        return os.getenv(self.api_env, '')
```
parente: Make sure validation is done against the env var flavor too.

kernel_gateway/gatewayapp.py
```diff
@@ -198,12 +206,23 @@ def init_configurables(self):
         if self.seed_uri is not None:
             self.seed_notebook = self._load_notebook(self.seed_uri)
 
-        self.kernel_manager = SeedingMappingKernelManager(
-            parent=self,
-            log=self.log,
-            connection_dir=self.runtime_dir,
-            kernel_spec_manager=self.kernel_spec_manager
-        )
+        if self.api:
```
parente: See above. No longer a boolean.

kernel_gateway/gatewayapp.py
```diff
@@ -198,12 +206,23 @@ def init_configurables(self):
...
+            with open(self.api) as nb_fh:
+                notebook = nbformat.read(nb_fh, 4)
```
parente: Why is this not using `_load_notebook` which supports both local and remote notebooks and does the "do I have a valid kernel" check?

kernel_gateway/gatewayapp.py
```diff
@@ -221,17 +240,31 @@ def init_webapp(self):
...
+        if self.api:
+            kernel_id = self.kernel_manager.start_kernel()
+            kernel_client = self.kernel_manager.get_kernel(kernel_id).client()
+            # discover the notebook endpoints and their implementations
+            endpoints = self.kernel_manager.endpoints()
+            sorted_endpoints = self.kernel_manager.sorted_endpoints()
+            # append tuples for the notebook's API endpoints
+            for uri in endpoints:
+                parameterized_path = parameterize_path(uri)
+                print('Registering uri {}'.format(parameterized_path))
```
parente: Leftover print. Probably do want logging here though so that the person bringing things up can see what's exposed.

kernel_gateway/gatewayapp.py
```diff
@@ -221,17 +240,31 @@ def init_webapp(self):
...
-            # Some handlers take args, so retain those in addition to the
-            # handler class ref
-            new_handler = tuple([pattern] + list(handler[1:]))
-            handlers.append(new_handler)
+        if self.api:
+            kernel_id = self.kernel_manager.start_kernel()
+            kernel_client = self.kernel_manager.get_kernel(kernel_id).client()
+            # discover the notebook endpoints and their implementations
+            endpoints = self.kernel_manager.endpoints()
+            sorted_endpoints = self.kernel_manager.sorted_endpoints()
```
parente: Should this be used?

kernel_gateway/services/kernels/manager.py
```diff
@@ -63,3 +65,108 @@ def start_kernel(self, *args, **kwargs):
...
+        kernel_id = super(MappingKernelManager, self).start_kernel(kernel_name=self.kernelspec, *args, **kwargs)
+        if kernel_id and self.source is not None:
+            # Only run source if the kernel matches
+            kernel = self.get_kernel(kernel_id)
+            if kernel.kernel_name == self.kernelspec:
+                # Connect to the kernel and pump in the content of the notebook
+                # before returning the kernel ID to the requesting client
+                client = kernel.client()
+                for code in self.source:
+                        # Execute every non-API code cell and wait for each to succeed or fail
```
parente: Indent fix.

kernel_gateway/services/kernels/manager.py
```diff
@@ -63,3 +65,108 @@ def start_kernel(self, *args, **kwargs):
...
+            self._source = None
+
+        return self._source
+
+    def start_kernel(self, *args, **kwargs):
+        '''
+        Starts a kernel and then optionally executes a list of code cells on it
+        before returning its ID.
+        '''
+        if self.kernelspec == 'r':
```
parente: Should probably refactor this into a little dictionary with a None entry for the default case.

kernel_gateway/services/notebooks/handlers.py
```diff
@@ -0,0 +1,88 @@
...
+            return
+
+        source_code = self.sources[self.request.method]
+        REQUEST = json.dumps({
+            'body' : parse_body(self.request.body),
+            'args' : parse_args(self.request.arguments),
+            'path' : self.path_kwargs
+        })
+        if self.kernel_name == 'r':
+            request_code = "REQUEST <- '"  + REQUEST + "'"
```
parente: I'd go with a delegation model here. Instead of a bunch of if/elses, just invoke a function `assignment_for_lang` and pass it the kernel name. Let that function decide what operator to return, probably using a little dictionary trick again.

kernel_gateway/services/notebooks/handlers.py
```diff
@@ -0,0 +1,88 @@
...
+        result, status = self._send_code(source_code)
+
+        # TODO: Will we need to add the ability to specify mime types?
+        self.set_header('Content-Type', 'text/plain')
+        if result is not None:
+            self.write(result)
+        self.set_status(status)
+        self.finish()
+
+    def get(self, **kwargs):
```
parente: Is there a tornadoism for defining one "handle all types of HTTP requests" method instead of each of these?

kernel_gateway/services/notebooks/handlers.py
```diff
@@ -0,0 +1,88 @@
...
+    def _send_code(self, code, block=True):
+        # execute the code and return the result and HTTP status code
+        self.kernel_client.execute(code)
+        result_found = False
+        iopub_message = None
+        result = None
+        status = 200
+        try:
+            while(result is None):
+                iopub_message = self.kernel_client.get_iopub_msg(block=block, timeout=self.execution_timeout)
```
parente: For now, block=True is fine. We need to think about if we want to support running multiple kernels behind the gateway and having the gateway act as a load balancer of sorts across them or not. Right now, we're assuming the kernel+gateway combo will be scaled out by a higher power.

kernel_gateway/services/notebooks/handlers.py
```diff
@@ -0,0 +1,88 @@
...
+
+        return result, status
+
+    def _handle_request(self):
+        if self.request.method not in self.sources:
+            self.set_status(405)
+            self.finish()
+            return
+
+        source_code = self.sources[self.request.method]
```
parente: This will 500 error if the request type is not found. Should try/except KeyError and raise a proper 405 Method Not Allowed instead in this case.

parente:
Some system tests would also be good. You can keep cramming them into `test_gatewayapp` for now and we can figure out how to refactor later.

kernel_gateway/services/notebooks/handlers.py
nitind: Nothing obvious that I can find, just mechanisms for pulling out common preprocessing and postprocessing code. We thought allowing for the different methods would be a plus, actually, and give us a clear way to declare a cell as one that should be exposed.

kernel_gateway/gatewayapp.py
nitind: Do we need a dedicated option for that plus another specifying the notebook whose APIs should be exposed, or just renaming and documenting this one for clarity?

parente: The two options are separate. The seed notebook makes sense under both API modes (and potentially new modes that will be added.)

kernel_gateway/services/notebooks/handlers.py
parente: Supporting the 4 methods is fine, I agree. I was just curious if there was a way in tornado to avoid implementing 4 essentially blank methods that all do the same thing.

kernel_gateway/gatewayapp.py
nitind: `_load_notebook` always reads `self.seed_uri` instead of the `uri` that's passed in, which looks more like a bug than something intentional now. Will fix that and use it.

parente: Ah, I see that you were actually passing the notebook as the API instead of a value indicating the *type* of API to expose. Yes. The `seed_uri` should point to the notebook. The `api` should indicate the type of API the gateway is making available.

kernel_gateway/gatewayapp.py
nitind: I don't follow.

parente: Sorry. `sorted_endpoints` appears to be an unused after it is assigned.

nitind: Ah, that's also addressed in last night's commit.

kernel_gateway/services/notebooks/handlers.py
parente: Nevermind. I see the 405 above it.

parente:
OK. I think this is MVP enough to put into master. I've opened some follow on issues. And we'll open more as we find bugs or potential improvements.

Nice work @nitind and @Lull3rSkat3r. I think people are going to have some fun with this once we make it easy to convert and deploy.

Issue #43, opened 2015-12-14
Ignore Whitespace In API Paths


Lull3rSkat3r:
When there is a space trailing a path or a path parameter at the end the endpoint registration will not register properly. Examples include:

```python
# GET /foo/:bar 
# GET /foo/:bar/quo 
```

parente:
Fixed by PR #44

PR #44, opened 2015-12-14
[Issue 43]: Fixed regex for path parameter extraction and trim endpoints

Lull3rSkat3r:
This PR removes all whitespace at the end of an API endpoint. This was causing `404` errors when making requests because the tornado handlers were being registered with spaces.


parente:
Ref #43

parente:
Is this WIP or ready to go in? LGTM.

Issue #49, opened 2015-12-15
kernel_name is not correctly set when pre spawning kernels


poplav:
When initializing and starting the kernel pool we are not correctly forwarding the kernel name on the start kernel operation.  This is evident in the [constructor](https://github.com/jupyter-incubator/kernel_gateway/blob/master/kernel_gateway/services/kernels/pool.py#L26).  Due to the kernel name not being set it defaults to python3 [ in multikernelmanager.start_kernel](https://github.com/jupyter/jupyter_client/blob/622f06d58b4671da4c585369d52ce11ba3737955/jupyter_client/multikernelmanager.py) and not the proper kernel name in the notebook.

poplav:
PR #51 will fix the issue.

PR #51, opened 2015-12-15
[Issue 49]: Fix kernel_name is not correctly set when pre spawning kernels

poplav:
(c) Copyright IBM Corp. 2015

### DIFF
--- a/kernel_gateway/tests/test_gatewayapp.py
+++ b/kernel_gateway/tests/test_gatewayapp.py
@@ -559,7 +559,7 @@ class TestSeedGatewayAppKernelLanguageSupport(TestGatewayAppBase):
             'parent_header': {},
             'channel': 'shell',
             'content': {
-                'code': 'print(this.s)',
+                'code': 'print this.s',
                 'silent': False,
                 'store_history': False,
                 'user_expressions' : {}
--- a/kernel_gateway/tests/test_gatewayapp.py
+++ b/kernel_gateway/tests/test_gatewayapp.py
@@ -5,6 +5,7 @@ import logging
 import unittest
 import os
 import sys
+import json
 
 from kernel_gateway.gatewayapp import KernelGatewayApp, ioloop
 from jupyter_client.kernelspec import NoSuchKernel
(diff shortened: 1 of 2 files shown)


parente:
Looks like the right stuff. We should get a test in for a non-Python3 notebook at least. The dev container has Python2 also. How about a test notebook written for Python2 and a test ensuring that the right kernel launches somehow?

kernel_gateway/tests/test_gatewayapp.py
```diff
@@ -524,6 +524,65 @@ def test_seed(self):
...
+    @gen_test
+    def test_seed_kernel_name_language_support(self):
+        '''Python2 Kernel should have the correct name'''
+        _, kernel_id = yield self.app.kernel_pool.acquire()
+        kernel = self.app.kernel_manager.get_kernel(kernel_id)
+        self.assertEqual(kernel.kernel_name,'python2')
+
+    @gen_test
+    def test_seed_language_support(self):
+        '''Python2 Kernel should have variables preseeded from notebook. Failures may be the result of networking problems.'''
```
Lull3rSkat3r: This second test is not doing anything python2 specific. You can change the print statement below to be:
```
print this.s
```
This would verify the kernel is running python2. If your intent was not to check that the code is python2 then I would remove this tests because it is testing the same thing as the test in `TestSeedGatewatApp`.

kernel_gateway/tests/test_gatewayapp.py
```diff
@@ -524,6 +524,68 @@ def test_seed(self):
...
         ws.close()
 
+class TestSeedGatewayAppKernelLanguageSupport(TestGatewayAppBase):
+    def setup_app(self):
+        self.app.prespawn_count = 1
+        self.app.seed_uri = os.path.join(RESOURCES,
+                                         'zen{}.ipynb'.format(sys.version_info.major))
+    @coroutine
+    def spawn_kernel(self):
+        ws = yield super(TestSeedGatewayAppKernelLanguageSupport, self).spawn_kernel(kernel_body="{{\"name\":\"python{}\"}}".format(sys.version_info.major))
```
parente: Nit. Might just want to use json.dumps() on a dict for clarity instead of building a JSON string manually.

parente:
Fixed #49

Issue #57, opened 2015-12-22
Request Code is not properly interpreted by the Julia REPL


poplav:
The issue arises when using the kernel gateway in notebook-http mode using a Julia kernel.

In the [kernel_gateway/handler](https://github.com/jupyter-incubator/kernel_gateway/blob/master/kernel_gateway/services/notebooks/handlers.py#L95), request code is being executed before source code, where the source code is dependent on the execution of the request code.  The Julia REPL treats strings differently than the other kernels like R and Python.  Due to Julia treating strings differently the request code will not be properly assigned resulting in errors in the source code.  Specifically, Julia does not support ``` ' ``` for strings.  An example of the issue is apparent in the following files/comparisons of the REPLs.

![julia_r_python_repl_json](https://cloud.githubusercontent.com/assets/2311898/11965783/92bbee9e-a8bc-11e5-977c-f17765c07095.png)

![julia_string](https://cloud.githubusercontent.com/assets/2311898/11965785/97d20ab2-a8bc-11e5-8284-9c33739bff9f.png)

parente:
Ah, yes. The JSON string assignment to REQUEST has to be made more language neutral or has to become a language map like the assignment operator is.

/cc @nitind @Lull3rSkat3r

poplav:
PR [#58](https://github.com/jupyter-incubator/kernel_gateway/pull/58) should resolve this issue.  I tested the branch against the [Julia scotch notebook API] (https://github.com/poplav/kernel_gateway_demos/blob/issue-4-port-scotch-api/scotch_demo/notebooks/scotch_api_julia.ipynb) I was developing and it looks to work while not breaking anything else.

PR #58, opened 2015-12-23
[Issue 57]: Change to using double-quotes for JSON strings

poplav:
(c) Copyright IBM Corp. 2015

### DIFF
--- a/kernel_gateway/services/notebooks/handlers.py
+++ b/kernel_gateway/services/notebooks/handlers.py
@@ -21,7 +21,7 @@ class NotebookAPIHandler(TokenAuthorizationMixin, CORSMixin, tornado.web.Request
     stream_messages = []
     error_message = None
     _assignment_statements = {'r': "REQUEST <- '{}'",
-        None : "REQUEST = '{}'"}
+        'julia-0.3': "REQUEST = \"{}\"", None: "REQUEST = '{}'"}
 
     def initialize(self, sources, kernel_pool, kernel_name):
         self.kernel_pool = kernel_pool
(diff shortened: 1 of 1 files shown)


kernel_gateway/services/notebooks/handlers.py
```diff
@@ -35,6 +35,11 @@ def _request_assignment_for_lang(self, kernel_name, expression):
             statement = self._assignment_statements[None]
         return statement.format(expression)
 
+    def _request_json_for_lang(self, kernel_name, expression):
+        if kernel_name == 'julia-0.3':
+            expression = expression.replace('\"', '\\"')
```
parente: Instead of patching for Julia specifically, I think we should fix this in general. I suspect the number of languages that use double-quotes for strings instead of single quotes is going to be higher. So let's get code in that uses the double-quote syntax by default and properly escapes the JSON string quotes with backslashes.

kernel_gateway/services/notebooks/handlers.py
```diff
@@ -20,19 +20,15 @@ class NotebookAPIHandler(TokenAuthorizationMixin, CORSMixin, tornado.web.Request
...
         self.sources = sources
         self.kernel_name = kernel_name
 
-    def _request_assignment_for_lang(self, kernel_name, expression):
-        try:
-            statement = self._assignment_statements[kernel_name]
-        except KeyError:
-            statement = self._assignment_statements[None]
+    def _format_request(self, expression):
+        expression = expression.replace('\"', '\\"')
```
parente: A comment in here about why we're doing this escaping will probably help future contributors.

parente:
Fixed #57

PR #59, opened 2015-12-31
[Issue 48]: Added the optional /api/source endpoint to download the notebook source in http mode.

poplav:
(c) Copyright IBM Corp. 2015

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -169,12 +169,10 @@ class KernelGatewayApp(JupyterApp):
 
     allow_notebook_download_env = 'KG_ALLOW_NOTEBOOK_DOWNLOAD'
     allow_notebook_download = Bool(config=True,
-                   default_value=False,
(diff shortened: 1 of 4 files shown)


kernel_gateway/gatewayapp.py
```diff
@@ -238,6 +245,10 @@ def init_webapp(self):
         # Redefine handlers off the base_url path
         handlers = []
         if self.api == 'notebook-http':
+            #Register the NotebookDownloadHandler if configuration allows
+            if self.allow_notebook_download:
+                handlers.append((r'/api/source', NotebookDownloadHandler, {'path': self.seed_uri}))
```
parente: Check with @Lull3rSkat3r about `/api`. We might have decided somewhere in the swagger work that we'd use the `/_api` for magic endpoints that we could expose without interfering with the API exposed by the notebook itself.

poplav:
@parente, modified api to _api.  It should be ready for another look.

PR #67, opened 2016-01-06
[Issue 63]: Fix escape double quotes in request parsing.

poplav:
Remove all escaped quotes from the original request.  Once removed then escape all quotes in the request.  This issue arrises on subsequent requests, for example: "If-None-Match": "\"9a28a9262f954494a8de7442c63d6d0715ce0998\"", in the original request.  If the original escaped quotes are not removed the quotes will be escaped multiple times.  Tested across multiple kernels.
(c) Copyright IBM Corp. 2015


parente:
Can't do the removal in case other parts of the content contains `\\\\\\""""` or some such. Suggest double encoding.

Lull3rSkat3r:
Tested this in the browser and with Postman doing a `GET` and `POST` with escaped double quotes in headers and the body. All worked fine, only bit we need to add in are test cases.

parente:
See #68.

Issue #61, opened 2016-01-05
Show Tornado request logs in notebook-http mode


parente:
Configure the logger in the same manner as when it's in jupyter-websocket mode so that we can see the HTTP requests to the notebook-defined API.

PR #69, opened 2016-01-08
[issue: 61] Show Tornado request logs in notebook-http mode

nitind:
Enable Tornado's "pretty logging" when in notebook-http mode.  Debug logging is already present in services.notebooks.handlers'  __handle_request function.


parente:
Fixes #61.

PR #72, opened 2016-01-14
[issue: 71] In jupyter-websocket mode, base handler tries to use jinja to render an error

nitind:
Avoid the jinja usage in our kernel handler by writing errors through the base
tornado handler.
(c) Copyright IBM Corp. 2016

### DIFF
--- a/kernel_gateway/services/kernels/handlers.py
+++ b/kernel_gateway/services/kernels/handlers.py
@@ -3,6 +3,7 @@
 
 import tornado
 import notebook.services.kernels.handlers as notebook_handlers
+from notebook.base.handlers import IPythonHandler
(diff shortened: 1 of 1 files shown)


parente:
Tried this out, but the HTML page returned has a application/json content-type and does not render properly in the browser. Weird, right? I dug a little deeper and found that the methods on most/all of the services APIs in notebook are decorated with `@json_errors` (https://github.com/jupyter/notebook/blob/master/notebook/base/handlers.py#L442). The right thing to do for consistency,  then, is to decorate the methods in our subclass too, since we're overriding the originals. This sidesteps the jinja call plus returns JSON format errors to match what the rest of the API handlers do in the event of an error.

I'll open a new PR with the fix.

Issue #71, opened 2016-01-13
In jupyter-websocket mode, base handler tries to use jinja to render an error


parente:
The handler from notebook is overriding http://tornado.readthedocs.org/en/latest/web.html#tornado.web.RequestHandler.write_error and using jinja to render a pretty HTML page for a user. We should probably point it back to the most base class version (http://tornado.readthedocs.org/en/latest/_modules/tornado/web.html#RequestHandler.write_error) for consistency with notebook-http mode. That or consistently return no body or a JSON body like `{"status" : same_int_as_http_status_code, "message": "whatever we can"}`

```
ERROR:tornado.application:Uncaught exception in write_error
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/tornado/web.py", line 1443, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/usr/local/lib/python3.4/dist-packages/kernel_gateway/services/kernels/handlers.py", line 31, in get
    raise tornado.web.HTTPError(404, 'Not Found')
tornado.web.HTTPError: HTTP 404: Not Found (Not Found)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/tornado/web.py", line 974, in send_error
    self.write_error(status_code, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/notebook/base/handlers.py", line 364, in write_error
    html = self.render_template('%s.html' % status_code, **ns)
  File "/usr/local/lib/python3.4/dist-packages/notebook/base/handlers.py", line 302, in render_template
    template = self.get_template(name)
  File "/usr/local/lib/python3.4/dist-packages/notebook/base/handlers.py", line 298, in get_template
    return self.settings['jinja2_env'].get_template(name)
KeyError: 'jinja2_env'
```

nitind:
I'll take care of this.

odovad:
@nitind I think I encountered a similar issue : 

```
tornado.application - ERROR - Uncaught exception in write_error
Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/tornado/web.py", line 1681, in _execute
    result = self.prepare()
  File "/opt/conda/lib/python3.8/site-packages/jupyter_server/base/handlers.py", line 507, in prepare
    raise web.HTTPError(403)
tornado.web.HTTPError: HTTP 403: Forbidden

[...]

PR #73, opened 2016-01-15
Use json_errors decorator on kernel get/post

parente:
For consistency with notebook server. Also avoids a missing jinja exception.


parente:
Fixes #71

PR #83, opened 2016-01-21
Ensure base path is prefixed with forward slash.

jtyberg:
Fix case where base path/url is missing leading slash.

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -291,7 +291,7 @@ class KernelGatewayApp(JupyterApp):
                 default_base_handlers
             ):
                 # Create a new handler pattern rooted at the base_url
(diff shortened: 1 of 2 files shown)


kernel_gateway/gatewayapp.py
```diff
@@ -291,7 +291,7 @@ def init_webapp(self):
                 default_base_handlers
             ):
                 # Create a new handler pattern rooted at the base_url
-                pattern = url_path_join(self.base_url, handler[0])
+                pattern = url_path_join('/', self.base_url, handler[0])
```
parente: If it is already prefixed with a forward slash, this still works, right? (Does test case cover it?)

jtyberg: Yes, it still works if `base_url` has leading slash.  Those cases are covered by TestRelocatedGatewayApp.

PR #87, opened 2016-01-26
Implement /api/sessions in the kernel_gateway

bpburns:
A light weight implementation of /api/sessions without sqlite or the notebook's content manager.

We needed a way to associate kernel id's with a name.  When you are working on a notebook, the notebook path serves as an identifier.  For non notebook clients the notebook path can be used as the identifier for the application creating a session.

### DIFF
--- a/kernel_gateway/services/sessions/handlers.py
+++ b/kernel_gateway/services/sessions/handlers.py
@@ -1,11 +1,32 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import tornado
+from notebook.base.handlers import json_errors
 import notebook.services.sessions.handlers as notebook_handlers
 from ...mixins import TokenAuthorizationMixin, CORSMixin
 
+class SessionRootHandler(TokenAuthorizationMixin, 
+                        CORSMixin, 
+                        notebook_handlers.SessionRootHandler):
+    @json_errors
+    def get(self):
+        '''
+        Denies returning a list of running sessions  unless explicitly
+        enabled, instead returning a 403 error indicating that the list is 
+        permanently forbidden.
+        '''
+        if 'kg_list_kernels' not in self.settings or self.settings['kg_list_kernels'] != True:
+            raise tornado.web.HTTPError(403, 'Forbidden')
+        else:
+            super(SessionRootHandler, self).get()
+
 default_handlers = []
 for path, cls in notebook_handlers.default_handlers:
-    # Everything should have CORS and token auth
-    bases = (TokenAuthorizationMixin, CORSMixin, cls)
-    default_handlers.append((path, type(cls.__name__, bases, {})))
+    if cls.__name__ in globals():
+        # Use the same named class from here if it exists
+        default_handlers.append((path, globals()[cls.__name__]))
+    else:
+        # Everything should have CORS and token auth
+        bases = (TokenAuthorizationMixin, CORSMixin, cls)
+        default_handlers.append((path, type(cls.__name__, bases, {})))
--- a/kernel_gateway/services/sessions/handlers.py
+++ b/kernel_gateway/services/sessions/handlers.py
@@ -1,0 +1,11 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import notebook.services.sessions.handlers as notebook_handlers
+from ...mixins import TokenAuthorizationMixin, CORSMixin
+
+default_handlers = []
+for path, cls in notebook_handlers.default_handlers:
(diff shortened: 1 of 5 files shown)


kernel_gateway/services/sessions/sessionmanager.py
```diff
@@ -0,0 +1,171 @@
...
+from tornado import web
+
+from traitlets.config.configurable import LoggingConfigurable
+from ipython_genutils.py3compat import unicode_type
+
+class SessionManager(LoggingConfigurable):
+
+    def __init__(self, kernel_manager):
+        self.kernel_manager = kernel_manager
+        self._sessions = []
```
parente: Sessions are in-memory I see. Any concern about this filling up? Or is the thinking that the KG would be scaled out?

kernel_gateway/services/sessions/sessionmanager.py
```diff
@@ -0,0 +1,171 @@
...
+        self._columns = ['session_id', 'path', 'kernel_id']
+
+    def session_exists(self, path):
+        """Check to see if the session for a given notebook exists"""
+        return bool([item for item in self._sessions if item['path'] == path])
+
+    def new_session_id(self):
+        "Create a uuid for a new session"
```
parente: Style nit: triple quotes.

kernel_gateway/services/sessions/handlers.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import notebook.services.sessions.handlers as notebook_handlers
+from ...mixins import TokenAuthorizationMixin, CORSMixin
+
+default_handlers = []
+for path, cls in notebook_handlers.default_handlers:
+    # Everything should have CORS and token auth
+    bases = (TokenAuthorizationMixin, CORSMixin, cls)
```
parente: It's good that you're protecting the endpoints like we did in other cases. I wonder if the "list" action `GET /api/sessions` needs to become an option like `GET /api/kernels` did. That way, a public server could be stood up without different clients being able to "discover" one another simply by listing the sessions.

kernel_gateway/tests/test_jupyter_websocket.py
```diff
@@ -596,3 +596,47 @@ def test_seed_kernel_not_available(self):
         app = KernelGatewayApp()
         app.seed_uri = os.path.join(RESOURCES, 'unknown_kernel.ipynb')
         self.assertRaises(NoSuchKernel, app.init_configurables)
+
+class TestSessions(TestJupyterWebsocket):
```
parente: Test :trophy: for you!

parente:
Thanks for submitting this. One more comment in addition to the above: I wonder if the entire session API should be feature flagged or not. A purist perspective would say that sessions are specific to the notebook tool and don't belong here. And maybe that should be the default. But I grok the use case, and could see a `--KernelGatewayApp.sessions_api=True` or something being the workaround for "naming" kernels while retaining compatibility with the notebook server.

kernel_gateway/services/sessions/sessionmanager.py
bpburns: I see the KG as being scaled out.  The session manager would only be handling a few kernels at a time.

kernel_gateway/services/sessions/handlers.py
bpburns: That's probably a good idea since the data in sessions returns kernel ids.  It renders the protection around the kernels endpoint useless.  Maybe one protection setting covers both?

bpburns:
Regarding the session API being specific to the notebook tool.  Right now it is designed that way, it would be nice if it was a "client" based session api where "client" could be anything from notebooks to whatever alternative clients the gateway is serving.  Instead of {notebook: {path: "some.ipynb"}... it could become {application: {name: "some.ipynb",  type: "notebook"}... or something like that.

kernel_gateway/services/sessions/handlers.py
parente: I think we've got `--KernelGatewayApp.list_kernels=False` as the param name and default value at the moment. Maybe we deprecate that, make it cover both kernels and sessions listing, and add `--KernelGatewayApp.discovery=False` as a better name to cover both?

bpburns: Sounds like a plan

kernel_gateway/services/sessions/sessionmanager.py
parente: OK. Makes sense.

parente:
I updated the README and fiddled a bit with the tests in #90. Other than that, it looks good to go.

PR #97, opened 2016-02-09
Activity handler API

### LINKED ISSUE #96: Need ability to identify dormant kernels to quiesce them
In the context of a service that needs to efficiently manage jupyter-kernel-gateways for users, it is necessary to be able to monitor kernels to ascertain a sufficient period of inactivity, thereby triggering kernel shutdown. 

In the bigger picture, where a kernel-gateway is provisioned for each user, it is necessary to quiesce that user's kernel-gateway when they've "stepped away" from the API and left their kernels inactive for an extended period of time; gateway can be brought back up when they return. 

Since kernels can drive work outside of jupyter/gateway, there's more involved than just monitoring the kernel-gateway. But for the kernel-gateway, must be able to monitor for last real activity in a kernel, say timestamp. So, it would help to have an API that would return, for each kernel provisioned, the last time there was "activity" through the kernel.   Not sure about the best way to handle no kernels; if nothing returned, no way to determine how long since kernels were running unless state handled outside of the gateway.

Lull3rSkat3r:
Fixes #96

### DIFF
--- a/kernel_gateway/services/activity/__init__.py
+++ b/kernel_gateway/services/activity/__init__.py
@@ -1,0 +1,2 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
--- a/kernel_gateway/services/activity/handlers.py
+++ b/kernel_gateway/services/activity/handlers.py
@@ -1,0 +1,16 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import tornado.web
+import json
+from .manager import activity
+from ...mixins import TokenAuthorizationMixin, CORSMixin
+
+class ActivityHandler(TokenAuthorizationMixin, CORSMixin, tornado.web.RequestHandler):
+
+    def get(self, **kwargs):
+        if not self.settings.get('kg_list_kernels'):
+            raise tornado.web.HTTPError(403, 'Forbidden')
+        self.set_header('Content-Type', 'application/json')
+        self.set_status(200)
+        self.finish(json.dumps(activity.get()))
--- a/kernel_gateway/services/kernels/handlers.py
+++ b/kernel_gateway/services/kernels/handlers.py
@@ -5,9 +5,12 @@ import tornado
 from notebook.base.handlers import json_errors
 import notebook.services.kernels.handlers as notebook_handlers
 from ...mixins import TokenAuthorizationMixin, CORSMixin
+from ...services.activity.manager import activity, LAST_MESSAGE_TO_CLIENT, LAST_MESSAGE_TO_KERNEL, LAST_TIME_STATE_CHANGED, BUSY, CONNECTIONS, LAST_CLIENT_CONNECT, LAST_CLIENT_DISCONNECT
+from datetime import datetime
+import json
 
-class MainKernelHandler(TokenAuthorizationMixin, 
-                        CORSMixin, 
+class MainKernelHandler(TokenAuthorizationMixin,
+                        CORSMixin,
                         notebook_handlers.MainKernelHandler):
     @json_errors
     def post(self):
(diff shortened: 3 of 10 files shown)


kernel_gateway/services/activity/__init__.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (c) Jupyter Development Team.
```
parente: A bit weird to have this in the __init__ that gets ref'ed form the handler in the same package. How about an ActivityManager in keeping with the rest of the jupyter setup?

kernel_gateway/services/kernels/handlers.py
```diff
@@ -51,6 +54,54 @@ def get_json_body(self):
...
+    def on_close(self):
+        activity.publish(self.kernel_id, LAST_CLIENT_DISCONNECT, datetime.now().isoformat())
+        activity.decrement_activity(self.kernel_id, CONNECTIONS)
+        super(ZMQChannelsHandler, self).on_close()
+
+    def _on_zmq_reply(self, stream, msg_list):
+        msg_metadata = json.loads(msg_list[3].decode('UTF-8'))
+        # If the message coming back is a status message, we need to inspect it
+        if msg_metadata['msg_type'] == 'status':
+            msg_content = json.loads(msg_list[6].decode('UTF-8'))
```
parente: Parsing every message when the activity handler endpoints are not exposed will probably degredate performance. Can we add a quick if-check here and bypass all of this if the activity API is not enabled on startup?

kernel_gateway/services/activity/handlers.py
```diff
@@ -0,0 +1,16 @@
...
+import tornado.web
+import json
+from . import activity
+from ...mixins import TokenAuthorizationMixin, CORSMixin
+
+class ActivityHandler(TokenAuthorizationMixin, CORSMixin, tornado.web.RequestHandler):
+
+    def get(self, **kwargs):
+        if 'kg_list_kernels' not in self.settings or self.settings['kg_list_kernels'] != True:
+            raise tornado.web.HTTPError(403, 'Forbidden')
```
parente: Did we use 403 for the /api/kernels endpoint when listing is disabled? Or 404?

Lull3rSkat3r: I was referencing [this](https://github.com/jupyter-incubator/kernel_gateway/blob/master/kernel_gateway/services/kernels/handlers.py#L35) line of code, which I believe is the  `/api/kernels` endpoint.

parente:
Please update the README too to at least indicate that this API exists.

Lull3rSkat3r:
Updated the following:
* Moved and renamed `ActivityStore` to be `ActivityManager` in `manager.py`
* Updated README to list the api and how to enable it
* Added logic to not add activities if the `list_kernels` flag is not set.

kernel_gateway/services/activity/handlers.py
```diff
@@ -0,0 +1,16 @@
...
+# Distributed under the terms of the Modified BSD License.
+
+import tornado.web
+import json
+from .manager import activity
+from ...mixins import TokenAuthorizationMixin, CORSMixin
+
+class ActivityHandler(TokenAuthorizationMixin, CORSMixin, tornado.web.RequestHandler):
+
+    def get(self, **kwargs):
```
parente: Decorate this with `@json_errors`.

https://github.com/jupyter-incubator/kernel_gateway/blob/master/kernel_gateway/services/kernels/handlers.py#L27

parente:
I'm good with merging this after the addition of @json_errors.

kernel_gateway/services/kernels/handlers.py
```diff
@@ -51,6 +54,58 @@ def get_json_body(self):
...
+            activity.remove(kernel_id)
+        super(KernelHandler, self).delete(kernel_id)
+
+class ZMQChannelsHandler(TokenAuthorizationMixin,
+                        notebook_handlers.ZMQChannelsHandler):
+    '''This class listens for messages coming to and from the kernel to provide metrics
+    about kernel usage.
+    '''
+    def open(self, kernel_id):
+        if self.settings.get('kg_list_kernels', False) == True:
```
parente: `if self.settings.get('kg_list_kernels')` would be enough here. And in the other uses of this as well.

Lull3rSkat3r:
Push changes for the boolean checks, but the @json_errors annotation is throwing 500 errors. @parente is going look into the 500 errors.

parente:
I'll clean up the tests and @json_errors usage as part of #99 where I hope to have one final solution.

PR #106, opened 2016-02-19
Concatenate cells with repeat annotations

parente:
Support typical notebook development: lots of small cells, no need to join.

This is for notebook-http mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,4 @@ docs/_build/
 target/
 
 .DS_Store
+.ipynb_checkpoints/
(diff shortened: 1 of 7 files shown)


Lull3rSkat3r:
Do we want to include a system test with a notebook with multiple annotated cells?

parente:
Sure. Can if you think the unit test is not enough.

parente:
System test added. Cleaned up the test notebooks a bit too.

Lull3rSkat3r:
LGTM :shipit: :pig:

PR #114, opened 2016-03-04
[WIP] Troubleshoot Travis timeouts

willingc:


willingc:
@minrk Will you give a look at the 3 failures (each one is failing a different test but all three are failing in a similar way)?

I'm wondering if this is related to the random kernel issues that Fernando and Chris were seeing. Perhaps in jupyter_client, this line is timing out: https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/blocking/client.py#L28

parente:
There's two kinds of failures I've been seeing. One is kernels simply timing out, but that's been rarer until recently. The other is that all tests pass in travis, the test runner logs the tests as complete on stdout, but then the process / travis appears to hang indefinitely until travis times out and marks the build failed.

minrk:
I will see what I can find out. Thanks, @willingc!

willingc:
@parente I think the second type of failure is more likely Travis itself and not recognizing completion. This one is similar to what we had on the Notebook a month ago. 

The first type of failure is, I believe, some sort of state reporting that the kernel has timed out (even it hasn't) and then getting stuck in that state.

parente:
> I think the second type of failure is more likely Travis itself and not recognizing completion. This one is similar to what we had on the Notebook a month ago.

Agreed. To be clear, this is the "hang" I was asking @minrk about in gitter. I now recall i opened https://github.com/travis-ci/travis-ci/issues/5555 after seeing the same failure both here and against notebook.

> The first type of failure is, I believe, some sort of state reporting that the kernel has timed out (even it hasn't) and then getting stuck in that state.

This one is new and sporadic.

parente:
@willingc Also, to save you some time, here's all the crazy combinations I tried to fix the problem in the past:

https://github.com/jupyter-incubator/kernel_gateway/pull/91

willingc:
@parente Thanks for the additional detail. Do you have a rough time estimate (Month) when Travis started acting flaky.

As an FYI, Travis does have a linter for yaml files: http://yaml.travis-ci.org/

parente:
> Do you have a rough time estimate (Month) when Travis started acting flaky.

I'd say December timeframe for the travis hangs. One interesting thing I did not is that if I took out the 3.5 version from the travis.yml, then 3.4 started having the same hang-after-pass symptom. That's about when I threw up my hands last time.

I see you're fixing other nits along the way which is good. I'm doing the same along the way as I go through the docstring cleanup and opening defects about anything more than minor. Minor stuff I'm fixing along the way too.

parente:
> . I've gotten Travis to run all the tests without timing out. 

Cool. It looks like 3.5.1 in the last run on Travis still hit the hang. Did you figure something out that isn't in the PR yet?

> Now I have to figure out why some of the tests are failing.

The other failures about kernel timeouts and kernels dying are new. And never seem to happen using `make test` in the local dev environment.

I've only once hit an unexpected exception running locally. zmq blew up with a port conflict exception. I logged it over in https://github.com/jupyter-incubator/kernel_gateway/issues/85 and wondered if it could be related to the Travis failures. But never have I seen that particular exception on Travis.

parente:
And what really gets me is that the last merge to master here was all green:

https://travis-ci.org/jupyter-incubator/kernel_gateway/builds/113684900

Whereas this PR which effectively changes nothing is red all over. :/

parente:
>  On another note, I think I isolated two places in jupyter_client that may be causing the Runtime Errors in Travis.

Good. That's more than what I managed to figure out in the past.

On a related note, I installed the latest stable release of KG on my mac in a fresh Python 3.5 conda environment to see what errors I would hit, if any. Like in our docker dev environment, I see none of the runtime exceptions like we do on Travis. BUT, I do sporadically get the hang at the end of the successful test completion (I have to Ctrl-C to get nosetests to exit). This is the first time I've been able to reproduce this behavior outside of travis. 

I'm not sure how best to debug. Maybe attach gdb or pdb or http://stackoverflow.com/questions/132058/showing-the-stack-trace-from-a-running-python-application or ...

parente:
Using dtruss on my mac, I see the following pattern repeated endlessly after nose logs completion to stderr:

```
kevent(0x13, 0x0, 0x0)           = 0 0
socket(0x2, 0x1, 0x6)            = 6 0
fcntl(0x6, 0x2, 0x1)             = 0 0
fcntl(0x6, 0x3, 0x0)             = 2 0
fcntl(0x6, 0x4, 0x6)             = 0 0
connect(0x6, 0x101B637E8, 0x10)          = -1 Err#36
kevent(0x13, 0x700000101DD8, 0x1)                = 0 0
kevent(0x13, 0x0, 0x0)           = 1 0
kevent(0x13, 0x700000101E08, 0x1)                = 0 0
getsockopt(0x6, 0xFFFF, 0x1007)          = 0 0
close(0x6)               = 0 0
socket(0x2, 0x1, 0x6)            = 6 0
fcntl(0x6, 0x2, 0x1)             = 0 0
fcntl(0x6, 0x3, 0x0)             = 2 0
fcntl(0x6, 0x4, 0x6)             = 0 0
connect(0x6, 0x101C04278, 0x10)          = -1 Err#36
kevent(0x13, 0x700000101DD8, 0x1)                = 0 0
kevent(0x13, 0x0, 0x0)           = 1 0
kevent(0x13, 0x700000101E08, 0x1)                = 0 0
getsockopt(0x6, 0xFFFF, 0x1007)          = 0 0
close(0x6)               = 0 0
```

Still in some kind of loop requesting a socket perhaps? Maybe in a thread that prevents the main one from exiting?

parente:
This PR for jupyter_client 4.2: https://github.com/jupyter/jupyter_client/commit/6e2b10c8359e7016424f649bfa142b99e7e651cb?diff=split

changed the behavior of `wait_for_ready()` slightly. Instead of passing 0 down to the `socket.poll` function when called with no parameters, it's passing a default of 1 second timeout and repeatedly checking if the kernel `is_alive`. I suspect there's a race condition here. Perhaps on the first time through that loop, the kernel is not yet marked as `is_alive` so this outer loop exits prematurely with the "kernel died" message. Maybe `_pause` is still True from the `__init__` here https://github.com/jupyter/jupyter_client/blob/e4c5092c839ccc0690a3ba00d84607a353b31f57/jupyter_client/channels.py#L173 ? (@minrk, @takluyver)

Prior to this change, the code will still be sitting down in the `socket.poll` until the first message (the `kernel_info` most likely) was returned. We can't revert to the old behavior even by specifying the `wait_for_ready(timeout=X)` because a timeout of 1 sec is always used in the loop and the timeout parameter is used in a calc to break the loop.

A correlation worth noting: the docker image we use for dev is using jupyter_client==4.1.1 which does not have this change and does not experience the kernel has died errors.

willingc:
I think the root cause may be a flipped inequality sign in jupyter_client/blocking/client.py in this line https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/blocking/client.py#L39

willingc:
Filed a PR agains jupyter_client.

parente:
> I think the root cause may be a flipped inequality sign

Good catch. Totally missed that and started digging into the is_alive path.

parente:
Hah. I see in @minrk 's PR it's both `_paused` and the flipped sign.

Ref: jupyter_client#139

willingc:
@parente Very happy to close this ;-)

parente:
You and me both. Thanks for diving in. 

I'll keep an eye out once jupyter client 4.2.1 starts getting used. I suspect we still have the end of test hang as an unrelated problem.

willingc:
@parente I think the end of test hangs will likely be solved too since the test runner yesterday was running all of them at some point in my debugging.

parente:
No such luck. Hang still occurs with the fixed jupyter_client: https://travis-ci.org/jupyter-incubator/kernel_gateway/jobs/113997896

PR #118, opened 2016-03-07
Improve ActivityManager

### LINKED ISSUE #107: ActivityManager tuneup
* [x] Instantiate singleton in gatewayapp for consistency with all the other managers. Put it in the app settings and ref it from there, not via a module import.
* [x] No need for dummy_map. Just make the global defaults a dict.
* [x] `ignore` grows endlessly
* [x]  ~~Instead of checking kg_list_kernels everywhere that calls ActivityManager methods, just check it in the ActivityManager instance itself. (Easy once it gets instantiated in the gatewayapp and can get the settings.)~~
* [x] ~~Calls to the activity manager should probably happen *after* the action takes place in case it fails.~~

parente:
Fixes #107


Lull3rSkat3r:
LGTM :+1:

PR #115, opened 2016-03-05
Add docstrings

willingc:
@parente While debugging some of the Travis stuff, I added some docstrings while reading these files.

PR #122, opened 2016-03-08
Docstrings for all tests

parente:
Fixed a few test name collisions too

Finishes the fixes for #115

### DIFF
--- a/kernel_gateway/tests/services/activity/test_activity.py
+++ b/kernel_gateway/tests/services/activity/test_activity.py
@@ -1,5 +1,6 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+"""Tests for tracking client/kernel activities."""
 
 import unittest
 from kernel_gateway.services.activity.manager import ActivityManager, LAST_MESSAGE_TO_CLIENT, LAST_MESSAGE_TO_KERNEL, LAST_TIME_STATE_CHANGED, BUSY, CONNECTIONS, LAST_CLIENT_CONNECT, LAST_CLIENT_DISCONNECT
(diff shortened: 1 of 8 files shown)


Carreau:
You seem to like some strong formatting consistencies. Do you want to activate something like https://landscape.io/ on this repo, which run some linter on the code base ? 

I'll review that in more details, but as it seem mostly cleanup,

+1

parente:
Just following the recs from @willingc and @fperez about matching numpy syntax. I doubt as people contributes it'll be perfect forever. But, hey, starting on the right foot.

Not familiar with landscape.io. Are you using it on other jupyter projects?

Carreau:
> Not familiar with landscape.io. Are you using it on other jupyter projects?

No, but as you said, starting on the right foot. 
Also on IPython & Notebook landscape choke a lot on traitlets, and find a lot of errors where there is none.

willingc:
@Carreau landscape.io looks pretty interesting. If free for open source, it could be helpful.

parente:
I don't mind if you want to try it. But if it gives a lot of false positives, we'll probably want to turn it back off.

Carreau:
> I don't mind if you want to try it. But if it gives a lot of false positives, we'll probably want to turn it back off.

Ok, that would be easy, it does not see this repo, and the syncing with GitHub failed. So no landscape for now.

Carreau:
Actually it works. 

There is a couple of false positive: https://landscape.io/github/jupyter/kernel_gateway/ we can likely tweak it.

fperez:
Thanks so much @parente! 

But more importantly is the actual substance of having docstrings everywhere.  I really want this to become something we simply don't pass on, ever.  Projects like scikit-learn or sympy have set a great example of simply making good docs a necessary and inescapable part of the project's workflow, and we haven't.

Examples like this are a great step in the right direction.

ps - I'm not particularly picky about `'''` vs `"""`, though I prefer the latter and it's nice to see consistency :)

Issue #116, opened 2016-03-06
Don't rebuild the swagger spec on every access


parente:
The notebook-defined API is immutable so the spec should be too. Cache it in `SwaggerSpecHandler.initialize` after first build.

PR #123, opened 2016-03-08
[issue: 116] Don't rebuild the swagger spec on every access

nitind:
Cache the swagger output in a class variable and make sure it is cleared between tests.

(c) Copyright IBM Corp. 2016


nitind:
For https://github.com/jupyter/kernel_gateway/issues/116 .

parente:
:shipit:

Issue #128, opened 2016-03-11
Pass environment metadata to kernels on POST /api/kernels


parente:
Make it possible for clients to pass additional metadata through when creating kernels and have that metadata show up in the kernel environment. 

Use case: passing secrets to kernels from clients without having those secrets hardcoded at KG start or having to pass them via code execution on the websocket.

For example:

```
POST /api/kernels

{
  "name": "python3",
  "env": {
    "APIKEY" : "123456"
  }
}
```

The notebook API spec does not support the "env" key. I'm proposing extending it.

https://github.com/jupyter/notebook/blob/master/notebook/services/api/api.yaml#L454

If this isn't acceptable, it can also be implemented via headers which is less than ideal but keeps 100% compatibility with the notebook API spec.

rgbkrk:
Big fan of that. Especially since the token is created by the user at the start.

/cc @jdfreder

minrk:
If we did this, it should blacklist certain environment variables, such as PATH, PYTHONPATH, etc.

parente:
Agreed. Or how about whitelisting anything with a `KG_*` prefix?

parente:
Working on this. Note: needs jupyter_client 4.2+ for this fix: https://github.com/jupyter/jupyter_client/commit/c446df389977090359000b80dbe3b0c03d1302ac

PR #129, opened 2016-03-14
Simplify custom default kernel logic

parente:
No need for custom get_json_body, just set the default on the kernel manager instance. Noticed this while working on #128


Lull3rSkat3r:
LGTM :shipit:

PR #130, opened 2016-03-15
Close all test fixture ioloop file descriptors

### LINKED ISSUE #94: Travis stalls after tests succeed
See travis-ci/travis-ci#5555

parente:
Fixes #94 (?!)

### DIFF
--- a/kernel_gateway/tests/__init__.py
+++ b/kernel_gateway/tests/__init__.py
@@ -1,0 +1,11 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+from tornado import ioloop
+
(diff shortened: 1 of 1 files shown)


parente:
The first run on travis failed, but not with a hang, but rather with the other "socket in use" exception we've seen in the past. I've since restarted the test run on travis 10 or so times manually and not once hit the hang bug. I did hit the "address in use" problem a few times, however.

So, at a minimum, this change does not make anything worse and possibly fixes the hang completely. It is certainly a band aid over some underlying leak in the test fixture, but I don't think we should waste any more time on trying to chase it down. Other test problems still exist, but we can continue to address those over time.

PR #133, opened 2016-03-15
Support env vars in requests to spawn kernels

### LINKED ISSUE #128: Pass environment metadata to kernels on POST /api/kernels

parente:
* Hack around lack of access to the start_kernel method in the base class with an override
* Whitelist KERNEL_ prefixed vars, blacklist everything else
* Add test case

Fixes #128

### DIFF
--- a/kernel_gateway/services/kernels/handlers.py
+++ b/kernel_gateway/services/kernels/handlers.py
@@ -4,6 +4,7 @@
 
 import json
 import tornado
+from functools import partial
 from datetime import datetime
 import notebook.services.kernels.handlers as notebook_handlers
 from ...mixins import TokenAuthorizationMixin, CORSMixin, JSONErrorsMixin
@@ -20,7 +21,8 @@ class MainKernelHandler(TokenAuthorizationMixin,
     """
     def post(self):
         """Overrides the super class method to honor the max number of allowed
-        kernels configuration setting.
+        kernels configuration setting and to support custom kernel environment
+        variables for every request.
 
         Delegates the request to the super class implementation if no limit is
         set or if the maximum is not reached. Otherwise, responds with an error.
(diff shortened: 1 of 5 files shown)


parente:
@minrk @rgbkrk Would appreciate your thoughts on this. The `partial` in there is a complete hack, but it felt better than reimplementing the entire base class method just to pass additional args to `km.start_kernel`. 

Could be we want to improve this in the notebook server to make the `start_kernel` easier to customize. Could also be that this is a place where the dependency inversion of KG -> notebook is shining through.

kernel_gateway/services/kernels/handlers.py
```diff
@@ -38,7 +40,23 @@ def post(self):
...
+        model = self.get_json_body()
+        if model is not None and 'env' in model:
+            if not isinstance(model['env'], dict):
+                raise tornado.web.HTTPError(400)
+            # Whitelist to KERNEL_* args only
+            env = {key: value for key, value in model['env'].items() if key.startswith('KERNEL_')}
+            # No way to override the call to start_kernel on the kernel manager
+            # so do a temporary partial (ugh)
+            orig_start = self.kernel_manager.start_kernel
+            self.kernel_manager.start_kernel = partial(self.kernel_manager.start_kernel, env=env)
```
rgbkrk: I'm partial to partials, though that's my pseudo-FP hat being on.

parente:
The other elephant in the room here is about the API spec. This technically breaks compatibility with the notebook REST API spec, even though the `env` object is optional. How should we manage the spec and these proposals going forward? I don't think anyone wants to see the notebook and KG APIs grow far apart.

rgbkrk:
I'd love to explore the API with those interested during the in-person dev meeting, even if we go ahead with this.

minrk:
I think most proposals like this one should be considered as a testing ground for changes to the upstream API. For instance, I see no reason why something like this would be unique to KG, so pushing it upstream once it's ironed out here and we decide people like it would make sense. That's just my opinion, of course.

:+1: to chatting about this at the dev meeting.

parente:
Relevant: @Lull3rSkat3r is working on exposing a swagger spec under the same route as in the notebook server here for #135. One line of thinking goes, as long as the API is documented, then there's really no harm in the KG and NB server APIs diverging slightly / temporarily.

rgbkrk:
Does the code-quality/landscape check ever finish?

parente:
It has on all the other PRs. @Carreau set it up so I'm not sure I can kick it or whatever.

I'm not in a rush to merge this. Was just making a note for a chat next week.

parente:
Or we can ship it. Sure! :package:

rgbkrk:
:grin: Right after I clicked the green button I saw "not in a rush".

PR #138, opened 2016-03-29
Wrong super class in `super()`

Carreau:
I think, unless it's made on purpose ?


parente:
You would think, except now all the tests are failing. So either the test is garbage or we should have doc'ed why we were skipping the parent class. (I don't think we were on purpose.)

Carreau:
Hooray for tests.

parente:
So, um, I'm getting a test failure on master totally unrelated to this PR, but somehow the last merge to master was green on travis. :joy:

I'll dig in.

parente:
Fix text fixture problem in #139. Back to debugging the original problem ...

parente:
#141 fixes the test failure problems.

PR #137, opened 2016-03-23
[WIP] [Issue 131] - Implement kernel connection file reset if fails to start for the first time

poplav:
Fixes #131

Issue #131, opened 2016-03-15
Kernel launch failure when address is in use


parente:
Rarely when running the jupyter/kernel_gateway test suite, the tests fail with the following exception due to a race condition between two kernels using the same port:

```
Traceback (most recent call last):
  File "/opt/python/3.3.5/lib/python3.3/runpy.py", line 160, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/opt/python/3.3.5/lib/python3.3/runpy.py", line 73, in _run_code
    exec(code, run_globals)
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/ipykernel/__main__.py", line 3, in <module>
    app.launch_new_instance()
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/traitlets/config/application.py", line 595, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-122>", line 2, in initialize
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/traitlets/config/application.py", line 74, in catch_config_error
    return method(app, *args, **kwargs)
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/ipykernel/kernelapp.py", line 412, in initialize
    self.init_sockets()
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/ipykernel/kernelapp.py", line 245, in init_sockets
    self.init_iopub(context)
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/ipykernel/kernelapp.py", line 250, in init_iopub
    self.iopub_port = self._bind_socket(self.iopub_socket, self.iopub_port)
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/ipykernel/kernelapp.py", line 174, in _bind_socket
    s.bind("tcp://%s:%i" % (self.ip, port))
  File "zmq/backend/cython/socket.pyx", line 487, in zmq.backend.cython.socket.Socket.bind (zmq/backend/cython/socket.c:5156)
  File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc (zmq/backend/cython/socket.c:7535)
zmq.error.ZMQError: Address already in use
```

IIRC, the port assignment is requested by a client (e.g., via a connection file). Having the kernel retry on another port means the client won't be able to communicate with that kernel.

[...]

PR #141, opened 2016-03-30
Fix call to super in kernel manager

parente:
* Invoke parent class start_kernel, not grandparent
* Now adds kernel restarter callback (any sideeffect?)
* Requires coroutine decoration now

Supersedes #138

### DIFF
--- a/kernel_gateway/services/kernels/handlers.py
+++ b/kernel_gateway/services/kernels/handlers.py
@@ -54,11 +54,11 @@ class MainKernelHandler(TokenAuthorizationMixin,
             orig_start = self.kernel_manager.start_kernel
             self.kernel_manager.start_kernel = partial(self.kernel_manager.start_kernel, env=env)
             try:
-                super(MainKernelHandler, self).post()
+                yield super(MainKernelHandler, self).post()
             finally:
                 self.kernel_manager.start_kernel = orig_start
         else:
-            super(MainKernelHandler, self).post()
+            yield super(MainKernelHandler, self).post()
 
     def get(self):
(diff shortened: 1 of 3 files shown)


parente:
Want to see how this works on travis a few times. Might be the restarter logic also helps solve #131?

parente:
Does not help with #131 / #137 .

parente:
@Carreau I think this one is good to go.

kernel_gateway/services/kernels/handlers.py
```diff
@@ -19,6 +20,7 @@ class MainKernelHandler(TokenAuthorizationMixin,
     """Extends the notebook main kernel handler with token auth, CORS, and
     JSON errors.
     """
+    @gen.coroutine
     def post(self):
```
minrk: In order for this to work, you need to wait on the Future returned by `super().post`. Since that's the only async thing here, you can either:

1. don't wrap in `gen.coroutine` and `return super().post(...)`
2. wrap in `gen.coroutine` and `yield super()...`

parente: Interesting that all the tests passed and manual testing worked too. Maybe didn't matter because the `super` does the `finish()` on the connection?

At any rate, I'll update to do option 2.

parente: Added the yields.

minrk: Thanks! The reason it works now is probably because we don't do any async operations in the default implementation of these calls. They are made async to allow for slower implementations to be used.

PR #145, opened 2016-03-31
[Issue 92]: Added default decorator for config options

Lull3rSkat3r:


Lull3rSkat3r:
@parente, could use a review. I was never able to get the warnings printed when running via `make dev` or `make test`.

parente:
I closed / reopened to kick landscape.io check. Looks like their PR support is experimental and sometimes hangs.

parente:
`python -m unittest discover` is where I spotted all the deprecation warnings originally. Seem to be gone now from the KG code. Many still appear from jupyter_client which probably hasn't made the leap yet.

This LGTM.

PR #149, opened 2016-04-01
Enable Infinite Test Coverage

Lull3rSkat3r:
The IEEE defines that anything divided by 0 is [±infinity](https://en.wikipedia.org/wiki/IEEE_floating_point). And seeing as we wouldn't know if the value is positive or negative, this PR may fall into the category of [Pascal's Wager](https://en.wikipedia.org/wiki/Pascal%27s_Wager). So, in the optimistic view of yielding a positively infinite amount of test coverage,  this PR removes all of the test code.

**Validation**
You can run the following snippet to verify the infinite coverage, because, you know, JavaScript:
```
node -p "'Current test coverage is: ' + `find kernel_gateway/ -type f -exec cat {} + | wc -l`/ `find kernel_gateway/tests/ -type f -exec cat {} + | wc -l`"
```


parente:
What's most interesting about this is that nosetests reports 40% coverage even with all the tests removed.

Happy 4/1.

PR #147, opened 2016-04-01
[Issue #143] Doc tweaks and improvements

parente:
* Fixes #143
* make docs target
* Cleanup docs/environment.yml
* More badges

PR #150, opened 2016-04-04
Fix typo from PR #147

willingc:
Fixes 1 typo on #147

### DIFF
--- a/docs/source/devinstall.md
+++ b/docs/source/devinstall.md
@@ -1,6 +1,6 @@
 ## Development Workflow
 
-This document includes instructions for setting up Dockerized development environment for the Jupyter Kernel Gateway. It also includes common steps in the developer workflow such as running tests, building doc, updating specs, etc.
+This document includes instructions for setting up Dockerized development environment for the Jupyter Kernel Gateway. It also includes common steps in the developer workflow such as running tests, building docs, updating specs, etc.
 
 ### Prerequisites
 


willingc:
@parente I also made you a maintainer for these docs on RTD.

parente:
Thanks! We're ready to cut a 0.5.0 release of KG I think. Anything special we need to do for the RTD site? Or does it automatically see and start to build branches / tags?

willingc:
If the docs are up to date with version 0.5.0 from a content perspective, which they seem to be, the only thing that needs to be done is to create the release as you have done in the past. The only important thing to remember when making a release is to have up to date docs checked into master before making the release.

After the release we should check RTD to see if the release is being built properly and displayed as stable.

parente:
@willingc Looks like latest rebuilt fine, but stable just spits out the blank RTD template site even though the git commit SHA looks to be correct and pointing to the 0.5.x branch. Not sure what magic button I need to press to fix it.

willingc:
I will give it a look.

parente:
:tropical_drink:

willingc:
@parente It looks good to me. Perhaps it was just a caching thing. 

In general, the Project page as well as the builds page and versions page are useful to check to see if things are configured correctly. They are :)
screen shot 2016-04-04 at 12 27 23 pm

You can see on the versions page that stable is pointing to the correct SHA1 hash.
screen shot 2016-04-04 at 12 27 42 pm

parente:
Yep, I checked all those and was confused why I was still seeing the blank docs template. It was definitely caching: a force cache flush and refresh shows me the correct docs for stable. Thanks!

willingc:
Great :D

Issue #103, opened 2016-02-17
Proof point: See if notebook-http mode can be made a "plugin"


parente:
Right now, the code for the Jupyter Notebook APIs (`jupyter-websocket` mode) is mixed in with the code for the notebook-defined REST API (`notebook-http` mode). This was done for convenience during incubation to prove that KG could support other means (transports, APIs, and protocols) for working with  kernels. Interestingly enough, I had a conversation the other day that resulted in my thinking about how kernels might operate if hooked to a message queue (e.g., Rabbit MQ). Immediately, I wondered if such a setup could be made possible via the kernel gateway.

I certainly don't want to keep hacking support for additional kernel comm mechanisms into the KG code base. I wonder instead of an API could be defined to allow such things to be plugged-in easily. My hesitation is that not everything will so easily reuse the APIs and classes from the notebook server like the microservice approach did.

I'd like to take a pass at turning the notebook-http mode, the jupyter-websocket mode, and maybe even the message-queue idea into "plugins" (for lack of a better word) to the kernel gateway. If there's common ground, it'll improve the extensibility and maintainability of the project. If there's very little commonality beyond the boilerplate CLI, it may suggest each of these deserves to be in its own project.

Just writing this down. Not planning to jump on it right away.

----

From discussion with @nitind:

Clarifying the changes being implemented:

* [x] changing "API" parameter from a string to a fully qualified module name (PR #173)
* [x] moving the current jupyter-websocket and notebook-http functionality into their own modules, not buried in services/ (PR #173)
* [x] relocate functionality specific to either mode into their modules (PR #173)
* [x] update documentation to call "modes" personalities and describe the steps to implement one (e.g., the factory function, the package, how to pass it to KG)
* [x] relocate traitlets specific to one mode into its module (e.g. KG_LIST_KERNELS)

/cc @fperez @minrk

minrk:
:+1: thanks, @parente!

fperez:
+1, great writeup, thanks!

parente:
@nitind and I spoke about this face-to-face. I think it's worth taking a crack at this now since "all's quiet on the western front" at the moment. Basics we discussed:

* try the plug-in approach first, don't worry about separate repos and so on
* identify where we're doing if/else checks for mode and think about refactoring that into separate packages named after the mode (e.g., `jupyter_websocket`, `notebook_http`)
[...]

PR #161, opened 2016-05-25
[[WIP]] [Issue: 103] Proof point: See if notebook-http mode can be made a "plugin"

nitind:
NOT READY TO BE MERGED

Relocates notebook-http mode and jupyter-websocket setup into separate modules.
For https://github.com/jupyter/kernel_gateway/issues/103.
(c) Copyright IBM Corp. 2016


parente:
Nice first step.

Two thoughts as you continue ...

1. I wonder if it will be possible to totally separate the modules required for the two modes (and other future modes) from each other into packages. It looks like right now there are a few things shared. Mixins and base handlers are examples of things that probably should be shared and live in KG "core". The kernel pool concept, on the other hand, only came about to support notebook-http mode. The degenerate one that exists for jupyter-websocket mode is only there for API parity in a sense. That's a case where it might be possible to have that module only in the notebook mode package and remove the stubbed one from being used in the websocket package.

2. Originally we put everything into a `services` directory to mirror the structure of the notebook project. With the idea of plugging in your own personality, we'll need to break with having one big services directory. Maybe `services/` becomes the place where common code across personalities lives, while `plugins` or `personalities` or ... becomes the directory where the implementations that ship with KG live. How the `--KernelGatewayApp.api=a_personality_package` gets implemented later for the issue will dictate part of this.

parente:
In passing while working on other issues, I noted that even though there's not an explicit mode check in our custom kernel manager, there's definitely still notebook-http specific code like the use of APICellParser here https://github.com/jupyter/kernel_gateway/blob/master/kernel_gateway/services/kernels/manager.py#L85. This should be refactored as well. But we need to be careful: the logic of pre-populating kernel memory is also used in jupyter-websocket mode. It just has special case logic for executing only some cells in notebook-http mode.

Maybe the kernel manager also needs a delegation model? Or maybe each package provides its own kernel manager? How much code is redundant?

nitind:
The degenerate kernel pool would seem expendable, but that's still used for the prespawning of kernels in jupyter-websocket mode. The question of whether the tornado loop is itself something to break out is probably not in scope.

For the APIParser, it would be simpler to let the provide the kernel manager with a test function for `filter` determining which cells to seed.

I'm not married to the name "personality", btw.

nitind:
What's in there now:
- changed "API" parameter from a string to a fully qualified module name, with the module expected to implement a `create_personality` factory function (the previous `jupyter-websocket` and `notebook-http` values will continue to work along side the module names)
- moved the current `jupyter-websocket` and `notebook-http` functionality into their own modules, no longer buried in `services/`
- relocate functionality specific to either mode into their modules, like the API cell parser and swagger generation for `notebook-http` mode

Relocating traitlets unique to one personality will require more exploration in its own PR, assuming we figure out how to do it.

kernel_gateway/gatewayapp.py
```diff
@@ -221,6 +211,21 @@ def _api_changed(self, name, old, new):
...
         return os.getenv(self.allow_notebook_download_env, 'False') == 'True'
 
+    def _load_api_module(self, module_name):
+        '''Tries to import the given module name'''
+        api_module = None
+        # some compatibility allowances
+        if module_name == 'jupyter-websocket':
+            module_name = 'kernel_gateway.jupyter_websocket'
+        elif module_name == 'notebook-http':
+            module_name = 'kernel_gateway.notebook_http'
```
parente: Cool. Glad this was straightforward.

kernel_gateway/gatewayapp.py
```diff
@@ -221,6 +211,21 @@ def _api_changed(self, name, old, new):
...
+        try:
+            api_module = importlib.import_module(module_name)
+        except ImportError:
+            # a ValueError will be triggered in the caller
+            pass
```
parente: Better to let the ImportError bubble. It's possible the user has provided a valid package name, but that package cannot be imported for other reasons (e.g., it is throwing an exception because of a missing dependency).

kernel_gateway/gatewayapp.py
```diff
@@ -312,19 +320,23 @@ def init_configurables(self):
...
-            self.kernel_pool = KernelPool(
-                self.prespawn_count,
-                self.kernel_manager
-            )
+        
+        api_module = self._load_api_module(self.api)
+        func = getattr(api_module, 'create_personality', None)
+        if func is None:
+            self.log.critical('ERROR: the personality factory function was not found in the API module {}'.format(self.api))
+            self.exit(1)
```
parente: Probably better to raise a RuntimeError here with the message you're logging as critical. `sys.exit(1)` is harsh when testing.

kernel_gateway/gatewayapp.py
```diff
@@ -312,19 +320,23 @@ def init_configurables(self):
...
+        self.personality = func(self, self.kernel_manager.seed_kernelspec)
+
+        self.personality.init_configurables(self.log, {
+            'allow_notebook_download': self.allow_notebook_download,
+            'prespawn_count': self.prespawn_count,
+            'kernel_manager': self.kernel_manager,
+            'base_url': self.base_url,
+            'seed_uri': self.seed_uri})
```
parente: These might build up over time. How about just passing `self` and letting the personality access whatever it needs from the application instance?

kernel_gateway/jupyter_websocket/__init__.py
```diff
@@ -0,0 +1,65 @@
+from ..base.handlers import default_handlers as default_base_handlers
```
parente: Copy/paste the Jupyter copyright header to the top of the file from one of the others.

kernel_gateway/gatewayapp.py
```diff
@@ -312,19 +320,23 @@ def init_configurables(self):
...
-                self.kernel_manager
-            )
-        else:
-            self.kernel_pool = KernelPool(
-                self.prespawn_count,
-                self.kernel_manager
-            )
+        
+        api_module = self._load_api_module(self.api)
+        func = getattr(api_module, 'create_personality', None)
```
parente: I'm a little confused about the second arg, the `None`. When is this not None? Why pass it in if it's always None?

kernel_gateway/jupyter_websocket/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+from ..services.kernelspecs.handlers import default_handlers as default_kernelspec_handlers
+from ..services.sessions.handlers import default_handlers as default_session_handlers
+from .handlers import default_handlers as default_api_handlers
+from notebook.utils import url_path_join
+
+class JupyterWebsocketPersonality(object):
+    '''Personality for standard websocket functionality, registering
+    endpoints that are part of the Jupyter Kernel Gateway API
+    '''
+    def __init__(self, kernelspec):
```
parente: It seems odd to have `kernelspec` be the one and only constructor arg. I think I know why you're passing it for notebook http server, but I only see `None` ever passed when creating a personality, so maybe I'm missing how it's intended used. 

Still, it seems like there should be a more elegant way. Would passing the gatewayapp instance to the constructor and accessing its attributes help?

kernel_gateway/tests/services/cell/test_parser.py
```diff
@@ -5,7 +5,7 @@
 import unittest
 import re
 import sys
-from kernel_gateway.services.cell.parser import APICellParser
```
parente: We should refactor the test folders to following the renaming of the services folders? For example, this one should be in `tests/notebook_http/cell/test_parser.py` probably.

kernel_gateway/notebook_http/__init__.py
```diff
@@ -0,0 +1,89 @@
+from ..base.handlers import default_handlers as default_base_handlers
```
parente: Copy/paste the Jupyter copyright header to the top of the file from one of the others.

kernel_gateway/jupyter_websocket/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+        seeding a new kernel.'''
+        # seed all code cells
+        return True
+
+    def shutdown(self):
+        """Stop all kernels in the pool."""
+        self.kernel_pool.shutdown()
+
+def create_personality(gatewayapp, kernel_name):
+    return JupyterWebsocketPersonality(kernel_name)
```
parente: Where is `kernel_name` set to a non-`None` value?

parente:
This is looking good! 🎉 

I left comments inline. Some simple fixes, a couple stylistic things, and one part I'm confused about (the kernel_name arg). I think we need to clean-up the personality API a bit (e.g., only pass gateway app to the constructor, let the instance store it from there and use it in other methods). Other than that, I think it's nearly ready to go in.

I'll wait for you to comment back before I pull and try here.

kernel_gateway/gatewayapp.py
nitind: Good point.

kernel_gateway/gatewayapp.py
nitind: It's the default return value if not found, not an arg being passed during invocation. Like the possible ImportError above, it's simpler to let it fail as an AttributeError with the reported line showing what was not found than to intentionally do anything more elaborate...right?

kernel_gateway/jupyter_websocket/__init__.py
nitind: It can be used to solve the same problem cropping up during tests, yes. Even waiting to instantiate the SessionManager until *after* calling `self.personality.init_configurables()` still leaves a race condition in the tests:
```
tornado.application: ERROR: Future <tornado.concurrent.Future object at 0x7feb145ebb00> exception was never retrieved: Traceback (most recent call last):
  File "/opt/conda/lib/python3.5/site-packages/tornado/gen.py", line 1017, in run
    yielded = self.gen.send(value)
  File "/srv/kernel_gateway/kernel_gateway/services/kernels/manager.py", line 82, in start_kernel
    if self.parent.personality.should_seed_cell(self.seed_kernelspec, code):
  File "/srv/kernel_gateway/kernel_gateway/notebook_http/__init__.py", line 84, in should_seed_cell
    return (not self.api_parser.is_api_cell(code) and not self.api_parser.is_api_response_cell(code))
AttributeError: 'NotebookHTTPPersonality' object has no attribute 'api_parser'
```

kernel_gateway/jupyter_websocket/__init__.py
nitind: JupyterWebsocketPersonality doesn't make use of it directly. The kernel manager uses the value from the gatewayapp itself.

kernel_gateway/gatewayapp.py
parente: That's what happens when I do code review at night. Somehow I missed the `getattr` and read it as invoking `creating_personality`. 👀 

Yes. I'd let the AttributeError bubble. If they've passed a valid package that doesn't have that function, let's fail fast and ugly.

nitind:
Closing in favor of cleaner merge from PR https://github.com/jupyter/kernel_gateway/pull/171 .

PR #171, opened 2016-06-16
[Issue: 103] Proof point: See if notebook-http mode can be made a "plugin"

nitind:
For issue https://github.com/jupyter/kernel_gateway/issues/103
(c) Copyright IBM Corp. 2016


nitind:
Replaces https://github.com/jupyter/kernel_gateway/pull/161

parente:
It's hard to see what comments from the last PR are addressed in this new one, which aren't, and what else might have changed over all. Can you summarize?

kernel_gateway/gatewayapp.py
```diff
@@ -221,6 +211,17 @@ def _api_changed(self, name, old, new):
...
 
+    def _load_api_module(self, module_name):
+        '''Tries to import the given module name'''
+        api_module = None
+        # some compatibility allowances
+        if module_name == 'jupyter-websocket':
+            module_name = 'kernel_gateway.jupyter_websocket'
+        elif module_name == 'notebook-http':
+            module_name = 'kernel_gateway.notebook_http'
+        api_module = importlib.import_module(module_name)
```
parente: Instead of hardcoding these, how about:

```
try:
  api_module = importlib.import_module(module_name)
except ImportError:
  api_module = importlib.import_module('kernel_gateway.'+module_name)
```

It's a cost at startup only, so the initial miss for the two built-in cases isn't a huge deal.

kernel_gateway/jupyter_websocket/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+from notebook.utils import url_path_join
+
+class JupyterWebsocketPersonality(object):
+    """Personality for standard websocket functionality, registering
+    endpoints that are part of the Jupyter Kernel Gateway API
+    """
+    def __init__(self, gatewayapp):
+        pass
+
+    def init_configurables(self, gatewayapp):
```
parente: This API is looking better. One more comment on consistency. The constructor gets a gatewayapp instance. `init_configurables` also gets it. The other methods do not.

How about just passing it to the constructor and letting the instance save a ref?

This is a bit of forward planning too. We'll want to make these LoggingConfigurables so that we can use traitlets to configure them. The convention for passing a ref like app through is to use the `parent` argument. Maybe we should just inherit now and adopt that convention so we don't have to change API when we switch over. It should be simple:

1. Have the personality classes inherit from LoggingConfigurable. 
2. Pass the gatewayapp instance to the personality constructor via a `parent=gatewayapp` keyword arg.
3. Have the constructor of the personality class call the ancestor constructors passing *args and **kwargs.
4. Change any code that needs the app ref to use `self.parent`.

Have a look at how ActivityManager's base class and constructor are definedhttps://github.com/jupyter/kernel_gateway/blob/master/kernel_gateway/services/activity/manager.py#L44 and how an instance is instantiated https://github.com/jupyter/kernel_gateway/blob/master/kernel_gateway/gatewayapp.py#L297. (Might as well pass the `log` too while you're at it.)

kernel_gateway/gatewayapp.py
nitind: Generally sounds better, but wouldn't it cause reported ImportErrors for properly missing modules to always be one that favors the 'kernel_gateway.'+ name?

parente: Yes. Good point. We'd still get the import error, but the exception log would be useless for tracking it down.

Let's keep it the way you've got it with the if-else for now for that reason.

nitind:
Rebased in https://github.com/jupyter/kernel_gateway/pull/173 , with just the specifics of the personality constructors possibly in need of more guidance.

PR #172, opened 2016-06-16
Switch 402 to 403 for kernel limit status

### LINKED ISSUE #158: max kernels breached yields 402 -- Payment required ... but i ain't payin' !
Officially, [w3](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) says:

```
10.4.3 402 Payment Required

This code is reserved for future use.
```

Seems like a tad harsh of a punishment for breaching this max ;-) Not to mention, this really isn't officially a status code in play at this time.

So how about this as an alternative:

```
10.4.4 403 Forbidden

The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated. If the request method was not HEAD and the server wishes to make public why the request has not been fulfilled, it SHOULD describe the reason for the refusal in the entity. If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead.
```

This seems appropriate because, indeed, kernel-gateway cannot fulfill the request. Further, the response can include the specific reason. This will make it crystal clear to the caller of the API.

parente:
Fixes #158

### DIFF
--- a/kernel_gateway/services/api/swagger.json
+++ b/kernel_gateway/services/api/swagger.json
@@ -178,7 +178,7 @@
                             }
                         }
                     },
-                    "402": {
+                    "403": {
                         "description": "The maximum number of kernels have been created.",
                         "schema": {
                             "$ref": "#/definitions/Error"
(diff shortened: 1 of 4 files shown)


jtyberg:
Makes sense.

PR #180, opened 2016-07-19
[WIP] Add support for additional API cell parsers

nitind:
Allow notebook-http to support alternate ways of marking cells as API endpoints. Add one that supports Swagger, and uses the same to help generate a more detailed API spec.

(c) Copyright IBM Corp. 2016


nitind:
Customizations for responses at runtime still needs to be through through. For now, it's the same as the existing simple API parser.

nitind:
Closing to resolve conflicts

PR #182, opened 2016-07-20
Add command line aliases for common opts

parente:
Reduce the number of times folks need to type KernelGatewayApp.*


jtyberg:
Yay, less typing!

PR #184, opened 2016-07-20
Option to serve static files under /public

parente:
For notebook-http mode only. Useful in cases where the notebook API wants
to refer to public static files without bothering the kernel to read them
and serve them itself (e.g., generated images).

### DIFF
--- a/kernel_gateway/notebook_http/__init__.py
+++ b/kernel_gateway/notebook_http/__init__.py
@@ -3,23 +3,20 @@
 """Notebook HTTP personality for the Kernel Gateway"""
 
 import os
+import tornado
 from ..base.handlers import default_handlers as default_base_handlers
 from ..services.kernels.pool import ManagedKernelPool
 from .cell.parser import APICellParser
 from .swagger.handlers import SwaggerSpecHandler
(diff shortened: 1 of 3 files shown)


parente:
Looking for thoughts on this one. I've hit use cases where I want to serve up some static content from my notebook API. To do so today, I've got to define an end point in the notebook, get the kernel to load the content, and then print it back out. This is slow, requires effort on the notebook author, and turns the kernel into a static file server. 

Tornado is much better at serving static files. This option makes it possible. But it definitely moves the mode away from being purely about APIs.

parente:
Example of what you can do with this:

I wrote a [single HTML file](https://github.com/parente/kernel_gateway_demos/blob/polymer-scotch/scotch_demo/static/index.html) containing web components that uses the scotch demo API from kernel_gateway_demos. I tossed the index.html file into a folder and [configured the KG to expose it](https://github.com/parente/kernel_gateway_demos/blob/polymer-scotch/scotch_demo/docker-compose.yml#L29) as `/public/index.html` using the new option in this PR. The result:

kg-static

lbustelo:
sweet!

PR #186, opened 2016-08-02
Add support for additional API cell parsers; Swagger implementation

nitind:
Allow notebook-http to support alternate ways of marking cells as API endpoints. Add one that supports Swagger, and uses the same to help output the actual API spec.

(c) Copyright IBM Corp. 2016

### DIFF
--- a/docs/source/config-options.md
+++ b/docs/source/config-options.md
@@ -160,6 +160,16 @@ NotebookHTTPPersonality options
     Default: False
     Optional API to download the notebook source code in notebook-http mode,
     defaults to not allow
+--NotebookHTTPPersonality.cell_parser=<Unicode>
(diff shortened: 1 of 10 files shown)


parente:
Looks like a legit failure on Travis: `AttributeError: 'TestSwaggerAPICellParser' object has no attribute 'assertLogs'`

nitind:
Makes sense. #assertLogs(logger, level) is new in 3.4...

parente:
LGTM. I'll merge it in. We can makes fixes and tweaks over time as it gets used. To that end, we'll probably want a small example of it over in kernel_gateway_demos. I'll open an issue there as a reminder. Anyone can work on it as time permits.

PR #188, opened 2016-08-09
Issue 187: dev image build failure from missing dependency

nitind:
Adds the enchant package to the dev/test image
(c) Copyright IBM Corp. 2016

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -2,6 +2,11 @@
 # Distributed under the terms of the Modified BSD License.
 
 FROM jupyter/pyspark-notebook:8015c88c4b11
+
+USER root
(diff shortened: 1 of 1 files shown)


nitind:
While it gets past the problem, I'm not sure this is the cleanest solution there could be.

parente:
It's just the doc that fails building, right? It's strange that it's never had a problem before and we haven't changed the docker image.

nitind:
Agreed. I suspect that dependencies in that environment, not being locked to specific versions, have changed to require this.

parente:
I don't think pyenchant is even being used. Nor some of the other installs. 

Opening a new PR that only includes the libraries used in the dashboards project doc build.

PR #214, opened 2017-01-22
Fix shutdown regression on interrupt

### LINKED ISSUE #206: Ctrl-C does not kill kernels.
I've noticed that hitting Ctrl-C on gateway does not kill kernels. There was an issue concerning this here: https://github.com/jupyter/kernel_gateway/issues/167 and then a pull request was merged here: https://github.com/jupyter/kernel_gateway/pull/168 However there was a regression here https://github.com/jupyter/kernel_gateway/commit/c85a4935fd7ba5383aa5a2b9be3b807435eae92f#diff-e12fea7b69535836a5f96101b1248316L474

Is it possible to bring back this change?

parente:
Fix #206

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -400,12 +400,12 @@ class KernelGatewayApp(JupyterApp):
     def init_http_server(self):
         """Initializes a HTTP server for the Tornado web application on the
         configured interface and port.
-        
+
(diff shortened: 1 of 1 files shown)


willingc:
Thanks @parente and @blink1073

parente:
Thanks for the look-see, both of you!

PR #216, opened 2017-01-22
Fix language comment syntax selection

### LINKED ISSUE #195: `notebook-http-mode` for different kernels
When trying to get the `notebook-http-mode` running with Apache Toree, I needed to change the code comment parsing and the creation of the `REQUEST` string. I changed the values inline, but we should probably discuss how to make this pluggable.

parente:
Use metadata.language_info.name property instead of metadata.kernelspec.name. Also remove code duplication between cell and swagger parsers.

Fixes #195


parente:
And another thank you.

PR #218, opened 2017-01-25
upgrade setuptools/pip before installing

minrk:
seems there's something wonky on Travis


parente:
Thanks @minrk. Not the same error, but @chuyqa reported some oddness using pip to install kernel gateway and notebook on python 2.7 recently too. Comments over here: https://github.com/jupyter/kernel_gateway/commit/c71091a6cd03bcc50aed6ee2172d9f7edb2014fc#commitcomment-20601516

PR #221, opened 2017-02-12
Allow config to whitelist kernel env vars

### LINKED ISSUE #200: Feature Req: Ability to pass in Spark Conf settings when creating Kernels
Today, the call to create a kernel does not allow to set Spark Conf setting. Therefore one cannot initialize user defined Spark conf properties. Here is an example 

 We provide some functionality via a lib that needs to be used both in spark-submit and interactive gateway. Here is how we use we have to use it from both

spark-submit  
 sparkSubmit.sh blah blah —conf spark.service.user.xxx.yyy.cred=demo blah blah blah

Kernel Gateway
System.setProperty("spark.service.user.xxx.yyy.cred","demo") 

The net is that you cannot add props to SparkContext after it is initialized i.e. downstream readers will not see it (some sparksql props are an exception)

this also means that our internal impl needs to handle system prop & conf and luckily in our case, our code only runs on driver so system.setProperty() will work. If the code needs to run on the executor, the system props would not get propagated to the executor and the user is blocked.

Therefore i think kernel creation in kernel gateway API, should allow to set user-specific sparkConf args

parente:
Fixes #200

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -428,6 +428,7 @@ class KernelGatewayApp(JupyterApp):
             kg_max_age=self.max_age,
             kg_max_kernels=self.max_kernels,
             kg_api=self.api,
+            kg_personality=self.personality,
             # Also set the allow_origin setting used by notebook so that the
             # check_origin method used everywhere respects the value
             allow_origin=self.allow_origin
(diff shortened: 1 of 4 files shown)


parente:
@mariobriggs, this PR adds a config option which lets you specify what environment variables can be passed to kernels from `POST /api/kernels` requests. For example:

```bash
jupyter kernelgateway --JupyterWebsocketPersonality.env_whitelist='["SPARK_OPTS","OTHER_STUFF"]'
```

parente:
@blink1073 Thanks for having a look.

Issue #154, opened 2016-04-20
Update to match the spec change in notebook 5.0 for session metadata


parente:
https://github.com/jupyter/notebook/pull/1359

Update the swagger.yml and simple session mgr implementation here to match.

parente:
This was fixed in #230.

PR #230, opened 2017-03-07
Fix #227: Switch to notebook 5 activity API

### LINKED ISSUE #227: Remove activity API in favor of the one in notebook 5.0
See https://github.com/jupyter/kernel_gateway/issues/226#issuecomment-280925032

parente:
* Update swagger
* Update dependencies
* Remove KG-specific API and tests


parente:
Also fixes #154, session metadata change for notebook 5.

parente:
Tests failing on travis on another session deserialize call. Probably have another spot n the code (or tests?) where we need a separate session instance.

parente:
Oddly enough, the tests all pass on my Mac. The exception seems to be coming from a tearDown method in the tests. I'll try to reproduce on my linux VM. It could also be that the setup on Travis is pulling different versions than what gets pulled in the dev setup.

rolweber:
The error message says:
```
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/notebook/services/kernels/kernelmanager.py", line 220, in record_activity
    msg = kernel.session.deserialize(fed_msg_list)
  File "/home/travis/virtualenv/python3.3.5/lib/python3.3/site-packages/jupyter_client/session.py", line 892, in deserialize
    raise ValueError("Duplicate Signature: %r" % signature)
nose.proxy.ValueError: Duplicate Signature: b'b8dcc527ffba3275ed1156892645bf258e74137c1e3de398cfd74da31d6bfc22'
```

And it's triggered here: 
https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/session.py#L892

Looks like a problem with signatures of Zero MQ messages. Maybe the Travis environment has a buggy version of Zero MQ or supporting libs?  
The jupyter_client keeps a history of message digests and checks for duplicates. Since the error occurs for all tested Python versions, it's not just pure chance that there's a duplicate. Either the check for duplicates is invalid, because there might be identical messages with identical signatures. Or there's something broken with the random or sequence numbers that would make messages and their signatures unique. Or the signature algorithm itself could be broken.

rolweber:
Or are there tests that use hard-coded messages to stub an actual kernel?

parente:
I was able to confirm that this also happens on my Linux VM.

minrk:
digest collision is due to the instantiation of `kernel.client()`, which shares `kernel.session` with the activity watcher. Neither the activity watcher (https://github.com/jupyter/notebook/pull/2398) nor newly instantiated clients (https://github.com/jupyter/jupyter_client/issues/252) should really be sharing the KernelManager's Session object. But in the meantime, you can workaround the issue by cloning the Session on the Client as well.

minrk:
I added the session clone here and tests are passing now.

parente:
I'll take one more pass through this PR by end of week. I think I had some junk in place for pre-releases and want to make sure all that's gone before merging.

parente:
OK. This should be set for review now. @rolweber would you care to take a look?

Issue #211, opened 2017-01-06
Support notebook 4.3 token generation (or equivalent)


parente:

parente:
The tokens in notebook as for one-time use, with the traditional cookie serving as the proof of authentication thereafter. That won't work for KG use cases where a cookie jar may not be available.

I think having an option to generate a secure token if one is not provided, and accepting that token as a URL parameter, especially on web socket connections, is the way to go.

rolweber:
I've just been trying to connect to a kernel gateway with auth token through a node.js program using the jupyterlab/services client. I can list kernel specs and running kernels. But when the web sockets come into play, after starting a kernel or when trying to execute code, things fail. The client reports loss of connection, and the kernel gateway a write error. I assume that this is because the kernel gateway doesn't understand the token being passed as a URL parameter to the web socket?

client output:
```
Starting websocket ws://localhost:8888/api/kernels/9e5be78e-91eb-498f-865a-e5fde5ce192f
connected
Kernel: reconnecting (9e5be78e-91eb-498f-865a-e5fde5ce192f)
Connection lost, reconnecting in 1 seconds.
Starting websocket ws://localhost:8888/api/kernels/9e5be78e-91eb-498f-865a-e5fde5ce192f
Connection lost, reconnecting in 2 seconds.
```

kernel gateway output:
```
[W 170303 09:21:07 web:1971] 401 GET /api/kernels/9e5be78e-91eb-498f-865a-e5fde5ce192f/channels?session_id=ce11007d55f7c30eff0d5caee8f97cca&token=Oh8Fifteen (172.17.0.1) 3.75ms
[KernelGatewayApp] Initializing websocket connection /api/kernels/9e5be78e-91eb-498f-865a-e5fde5ce192f/channels
[KernelGatewayApp] Using contents: services/contents
[E 170303 09:21:08 web:996] Uncaught exception in write_error
    Traceback (most recent call last):
      File "/opt/conda/envs/jupyter/lib/python3.5/site-packages/tornado/web.py", line 994, in send_error
        self.write_error(status_code, **kwargs)
      File "/opt/conda/envs/jupyter/lib/python3.5/site-packages/notebook/base/handlers.py", line 410, in write_error
        html = self.render_template('%s.html' % status_code, **ns)
      File "/opt/conda/envs/jupyter/lib/python3.5/site-packages/notebook/base/handlers.py", line 343, in render_template
        template = self.get_template(name)
[...]

PR #234, opened 2017-03-08
Websocket auth with token as query param

### LINKED ISSUE #233: Support token auth via query parameter
As discussed in #211, token auth for web sockets requires the token to be passed as a query parameter rather than an `Authorization` header. This is independent of the token generation logic mentioned in #211,  and compatible with the 1.x API.

rolweber:
Fixes #233 for version 2.0:
- unit tests for auth mixin
- code changes to auth mixin
- updated swagger
- fixed swagger link in docs

### DIFF
--- a/kernel_gateway/mixins.py
+++ b/kernel_gateway/mixins.py
@@ -59,7 +59,7 @@ class TokenAuthorizationMixin(object):
         server_token = self.settings.get('kg_auth_token')
         if server_token:
             client_token = self.get_argument('token', None)
-            if client_token == None:
+            if client_token is None:
                 client_token = self.request.headers.get('Authorization')
                 if client_token and client_token.startswith(self.header_prefix):
                     client_token = client_token[self.header_prefix_len:]
--- a/kernel_gateway/mixins.py
+++ b/kernel_gateway/mixins.py
@@ -39,6 +39,9 @@ class TokenAuthorizationMixin(object):
     """Mixes token auth into tornado.web.RequestHandlers and
     tornado.websocket.WebsocketHandlers.
     """
+    header_prefix = "token "
+    header_prefix_len = len(header_prefix)
+
     def prepare(self):
         """Ensures the correct auth token is present, either as a parameter
         `token=<value>` or as a header `Authorization: token <value>`.
@@ -55,11 +58,11 @@ class TokenAuthorizationMixin(object):
         """
         server_token = self.settings.get('kg_auth_token')
         if server_token:
-            client_token = self.get_argument('token', '')
-            if client_token == '':
+            client_token = self.get_argument('token', None)
+            if client_token == None:
                 client_token = self.request.headers.get('Authorization')
-                if client_token and client_token.startswith('token '):
-                    client_token = client_token[len('token '):]
+                if client_token and client_token.startswith(self.header_prefix):
+                    client_token = client_token[self.header_prefix_len:]
                 else:
                     client_token = None
             if client_token != server_token:
--- a/kernel_gateway/mixins.py
+++ b/kernel_gateway/mixins.py
@@ -40,10 +40,11 @@ class TokenAuthorizationMixin(object):
     tornado.websocket.WebsocketHandlers.
     """
     def prepare(self):
-        """Ensures the correct `Authorization: token <value>` is present in
-        the request's header if an auth token is configured.
+        """Ensures the correct auth token is present, either as a parameter
+        `token=<value>` or as a header `Authorization: token <value>`.
+        Does nothing unless an auth token is configured in kg_auth_token.
 
-        If kg_auth_token is set and the token is not in the header, responds
+        If kg_auth_token is set and the token is not present, responds
         with 401 Unauthorized.
 
         Notes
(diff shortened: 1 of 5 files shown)


rolweber:
I used `unittest.mock`, which is a Python 3 feature.
There's a backport to Python 2.7, but it uses a different package name.

parente:
Thanks for submitting this @rolweber. I'll have a pass through it soon. 

@minrk informed me yesterday that there is a long-term token capability in notebook 4.3+, not just the one time use token I noted in #211. I'm not sure if you found that or implemented your own (I haven't looked at the PR in depth yet), but it's something we should use if we can.

rolweber:
@parente: Thanks.

My PR doesn't generate tokens, it just verifies them. The difference is that it not only looks at the Authorization header, but also at query params. The token still has to be configured via KG_AUTH_TOKEN or similar.

kernel_gateway/mixins.py
```diff
@@ -54,8 +55,14 @@ def prepare(self):
         """
         server_token = self.settings.get('kg_auth_token')
         if server_token:
-            client_token = self.request.headers.get('Authorization')
-            if client_token != 'token %s' % server_token:
+            client_token = self.get_argument('token', '')
```
parente: Should probably default to None, not empty string.

kernel_gateway/mixins.py
```diff
@@ -54,8 +55,14 @@ def prepare(self):
...
+            if client_token == '':
+                client_token = self.request.headers.get('Authorization')
+                if client_token and client_token.startswith('token '):
+                    client_token = client_token[len('token '):]
```
parente: `len('token ')` should probably be computed once and stored a global in this module to avoid recompute it on every request.

kernel_gateway/tests/test_mixins.py
```diff
@@ -4,10 +4,125 @@
...
+    from unittest.mock import Mock, MagicMock
+except ImportError:
+    # Python 2.7: use backport
+    from mock import Mock, MagicMock
+
 from tornado import web
-from kernel_gateway.mixins import JSONErrorsMixin
+from kernel_gateway.mixins import TokenAuthorizationMixin, JSONErrorsMixin
+
+class SuperTokenAuthHandler(object):
```
parente: Is there a reason for defining this class just to have a `prepare()` instead of implementing prepare in `TestableTokenAuthHandler` which is the only subclass?

parente:
> My PR doesn't generate tokens, it just verifies them. The difference is that it not only looks at the Authorization header, but also at query params. 

@rolweber after reading through the jupyter/notebook master code, I still think the code in KG for generating and verifying tokens, plus the code in the PR here for checking query parameters for tokens, is redundant. https://github.com/jupyter/notebook/blob/68a514a29d0e79af66ce6dd75c87e991fb58126d/notebook/auth/login.py#L148, which feeds all of the `handlers.get_current_user` now supports tokens both in Authorization headers and token query parameters. Moreover, it's used on all endpoints by default including the Websocket endpoint. In the end, I think we want less code in the KG and to rely on as much of the code in notebook as possible.

kernel_gateway/tests/test_mixins.py
rolweber: I had to put it in a base class, because the tested method from the mixin calls super.prepare() in the good case. Maybe there's a more elegant way to do it, I still have to familiarize myself with Python.

rolweber:
@parente: You're probably right, but that's what issue #211 is about. I opened issue #234 with a smaller scope, because I'm trying to find my way into a code base I'm not familiar with, in a programming language I'm not familiar with. Issue #211 is simply out of my league at the moment.

I'll update the PR with your suggested improvements some time this week.

parente:
@rolweber Don't get me wrong: My notes above aren't meant to preclude merging this and then iterating on how to switch to what's in notebook server over time. I appreciate the PR you put together!

rolweber:
@parente: Thanks. The code is updated now.

rolweber:
ping?

kernel_gateway/mixins.py
```diff
@@ -55,11 +58,11 @@ def prepare(self):
         """
         server_token = self.settings.get('kg_auth_token')
         if server_token:
-            client_token = self.get_argument('token', '')
-            if client_token == '':
+            client_token = self.get_argument('token', None)
+            if client_token == None:
```
parente: `if client_token is None:`

parente:
@rolweber Missed the update from 7  days ago. Thanks for the ping. I had just one further piece of feedback, then we can merge.

rolweber:
@parente: Thanks! I'll try to find time for backporting this change to 1.2.x tomorrow.

parente:
@rolweber In it goes! 🍰  

Would you like to be added as a maintainer to this repo so that you can help review and merge PRs as well?

rolweber:
@parente: Great! It's been a while since I could last contribute to somebody else's OSS project :-)

Yes, it would be cool if I could become a maintainer here.

parente:
Whelp, it's your project now too. Welcome to the 🎉 .

Issue #232, opened 2017-03-08
Swagger: auth token


rolweber:
According to swagger.yaml, only `/api/kernelspecs` requires the auth token.
https://github.com/rolweber/kernel_gateway/blob/e897143cb8ff84a9f52819005524f168675966cb/kernel_gateway/jupyter_websocket/swagger.yaml#L72

I thought it is required for all calls, if configured? Should swagger list the token as a top-level `security` element? If there are only few calls which do not require the token, it would still be easier to exempt those than to add the token to many calls individually.

parente:
That was a copy/paste mistake, and PR #227 fixes it. It might be worth calling out in the top level security section how auth is handled globally.

PR #236, opened 2017-04-03
Update all swagger.json endpoints to support auth token header/query param

skbergam:


rolweber:
I thought Swagger allows to define global security constraints, so that you don't have to add the same setting to each endpoint? See also issue #232, and this commit from the other pending PR: https://github.com/jupyter/kernel_gateway/pull/230/commits/fcf1316ee399a422fcad7185e5ca9cc77454d8ae

skbergam:
@rolweber: maybe you're right, but I had to add these changes to enable the auth token on the other endpoints. could be that removing it from the one endpoint is all that was necessary for Swagger to infer it was meant to be global?

/cc @parente

rolweber:
I merged the other PR that had swagger updates. Could you please check if that resolves your problem?
If not, please update this PR. I suggest to use a top-level `security` element. According to the [spec](http://swagger.io/specification/), it is...

"A declaration of which security schemes are applied for the API as a whole."

parente:
I added the security requirements section in #244. I think we can close this PR. Thanks @skbergam.

Issue #237, opened 2017-04-05
JKG should catch SIGTERM and gracefully terminate


kevin-bates:
When invoked as a background process the general interrupt signal is not caught.  Moreover, SIGTERM is the defacto shutdown-my-process-and-cleanup signal.  It would be good to handle SIGTERM and invoke the stop action (which should be amended to shutdown active kernels).

kevin-bates:
See PR #238

PR #238, opened 2017-04-05
Issue #237: Handle SIGTERM and gracefully terminate

kevin-bates:
Add signal handler for termination requests and invoke stop sequence to shutdown kernels and stop application.

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -506,9 +506,9 @@ class KernelGatewayApp(JupyterApp):
         self.log.info('Jupyter Kernel Gateway at http{}://{}:{}'.format(
             's' if self.keyfile else '', self.ip, self.port
         ))
-        signal.signal(signal.SIGTERM, self._signal_stop)
-
         self.io_loop = ioloop.IOLoop.current()
+
+        signal.signal(signal.SIGTERM, self._signal_stop)
 
         try:
             self.io_loop.start()
@@ -532,6 +532,6 @@ class KernelGatewayApp(JupyterApp):
 
     def _signal_stop(self, sig, frame):
         self.log.info("Received signal to terminate.")
-        ioloop.IOLoop.current().stop()
+        self.io_loop.stop()
 
 launch_instance = KernelGatewayApp.launch_instance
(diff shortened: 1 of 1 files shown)


parente:
It looks like travis might be failing because of a backwards incompatibility with a second order dependency. The `ipykernel_launcher` appearing in the test log output is new.

I did confirm that notebook 4.x is still being installed, as desired for the moment. @minrk do you have any idea what else might have rev'ed and broken?

kernel_gateway/gatewayapp.py
```diff
@@ -528,5 +530,8 @@ def shutdown(self):
         """Stop all kernels in the pool."""
         self.personality.shutdown()
 
+    def _signal_stop(self, sig, frame):
+        self.log.info("Received signal to terminate.")
+        ioloop.IOLoop.current().stop()
```
rolweber: All other places use `self.io_loop`, it seems. But if that is used here, the signal handler should be installed after assigning `self.io_loop`.

minrk:
> @minrk do you have any idea what else might have rev'ed and broken?

There is a kernelspec installed where the executable is only `python` (this happens from wheels of ipykernel 4.6). This relies on resolution via PATH *in the subprocess*. When the env is overridden in KG, PATH is unset and thus the system default is applied. The result is that the kernel is launched with `/usr/bin/python` instead of the env's Python.

The fix should be for rest-api env to override the default env, not replace it. #239 should fix this.

kernel_gateway/gatewayapp.py
minrk: self.io_loop makes sense. `IOLoop.current()` uses thread-local storage, so may not find the right IO Loop, depending on whether the application and the event handler are running in the same thread.

parente:
> There is a kernelspec installed where the executable is only python (this happens from wheels of ipykernel 4.6).

Noting this for future reference mainly:

https://github.com/ipython/ipykernel/commit/654c7829c9b7e6984a0689d2b47c364a623436a1#diff-2eeaed663bd0d25b7e608891384b7298R93 appears to be the PR and line that sets the python executable path upon wheel install.

kernel_gateway/gatewayapp.py
kevin-bates: Excellent feedback - thank you for catching this.  I've tested the recommended changes and posted the commit.

parente:
Thanks @kevin-bates . @rolweber feel free to merge if everything looks good to you in the future, unless you want me to have a look too.

PR #239, opened 2017-04-06
POST env extends default env

minrk:
rather than starting from scratch with an empty env, which resets PATH among other things.

cf failures in #238

### DIFF
--- a/kernel_gateway/services/kernels/handlers.py
+++ b/kernel_gateway/services/kernels/handlers.py
@@ -3,6 +3,8 @@
 """Tornado handlers for kernel CRUD and communication."""
 
 import json
+import os
+
 import tornado
 import notebook.services.kernels.handlers as notebook_handlers
 from tornado import gen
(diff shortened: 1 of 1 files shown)


parente:
Thank you @minrk for tracking this down!

parente:
Dawned on me in writing the release notes for 2.0 that this change introduces a problem whereby secrets in the kernel gateway process environment (e.g., `KG_TOKEN`) can now be seen by kernels spawned by the KG. Either we should document this behavior or whitelist which env vars we want kernels to inherit (e.g., `PATH`).

rolweber:
I'd prefer the whitelist approach. I backported the change to 1.2.x and would rather not introduce a security-relevant change with a fixlevel.

rolweber:
For the records, see PRs #248 (master) and #250 (1.2.x).
We're only passing the PATH variable instead of the full environment there.

Issue #229, opened 2017-02-23
Documentation updates to clarify kernel_gateway's purpose


BigLep:
Hello,

Feel free to resolve/ignore, but I found that the docs under the kernel_gateway GitHub project weren't as helpful for communicating the problem that kernel_gateway is attempting to solve vs. http://blog.jupyter.org/2016/03/09/jupyter-kernel-gateway/ and https://github.com/jupyter-incubator/proposals/blob/master/kernel_gateway/proposal.md .  I'm only bringing this up because wouldn't want someone to prematurely discount/disregard the project.

Steve

parente:
Thanks for the feedback, @BigLep. If you have some time to contribute, we'll gladly take pull requests that improve the documentation. If not, we can try to work in some of the wording from the blog posts and proposal over time.

PR #246, opened 2017-05-11
#229 update project overview in readme and docs

rolweber:
Here's my suggestion to address issue #229. The text is identical for README and docs.
You can check the formatted README in my branch: https://github.com/rolweber/kernel_gateway/tree/229_doc_updates

@BigLep - is this clear enough?
@parente - is it correct? :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,14 +8,23 @@
 
 ## Overview
 
-The kernel gateway is a web server that supports different mechanisms for spawning and
-communicating with Jupyter kernels, such as:
+Jupyter Kernel Gateway is a web server that provides headless access to
+Jupyter kernels. Your application communicates with the kernels remotely,
+through REST calls and Websockets rather than ZeroMQ messages.
+There are no provisions for editing notebooks through the Kernel Gateway.
+The following operation modes, called personalities, are supported
+out of the box:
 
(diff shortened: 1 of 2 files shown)


parente:
@rolweber Thanks for tackling this!

The only thought I had was also about mentioning how the KG is using the Jupyter Notebook code base under the covers, not reimplementing everything from scratch. On one hand, a statement like that acts as indicator of code maturity here (a question that is sometimes asked). On the other, it's  a technical detail that might be more confusing to first time visitors than helpful.

rolweber:
@parente: Thanks for the suggestion. I found an unobtrusive way to mention it.

parente:
Very nice. I like it. Let's timebox leaving this unmerged for others to comment, say til Monday. We can plan on making the 2.0 release next week after this goes in.

BigLep:
Sorry for the delay.  The improvements look good to me.  My head has been out of this space, so don't have further comments for items in http://blog.jupyter.org/2016/03/09/jupyter-kernel-gateway/ and https://github.com/jupyter-incubator/proposals/blob/master/kernel_gateway/proposal.md that should make there way here.  Thanks!

parente:
Thanks all. Let's merge and improve over time.

PR #248, opened 2017-05-29
Only pass PATH from KG to new kernels

parente:
Keep the PATH env var when launching kernels. Do not provide the entire environment of the KG server. It might contain secrets that kernels should not see.


rolweber:
I'll need to backport that to 1.2.x as well.

Issue #257, opened 2017-09-04
Issue with using auth_token and CORS requests 


SimonBiggs:
When issuing a CORS request the browser sends a preflight OPTIONS http request. The browser doesn't put the token in this request. As a result kernel_gateway rejects the request.

This conflict appears to make CORS and auth_tokens not currently compatible. I've provided more details over at https://github.com/jupyterlab/jupyterlab/issues/2960

A solution might be to check if the request type is OPTIONS, if it is, then don't require the token to be provided.

I'd be happy to attempt a pull request. What are your thoughts?

Cheers,
Simon

parente:
Turning off the token check for OPTIONS sounds right to me. Feel free to submit a PR if you'd like. Thanks for reporting this.

rolweber:
Background information for those not familiar with the CORS protocol:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Overview
https://fetch.spec.whatwg.org/#cors-preflight-request

SimonBiggs:
Created a pull request over at:
https://github.com/jupyter/kernel_gateway/pull/258

parente:
Fixed by #258.

SimonBiggs:
@parente would a 2.0.1.dev0 release be able to be pushed to pypi? If I remember correctly users don't default to upgrading to dev releases, but I would be able to pin a version number in my own setup.py.

parente:
I'll work on a 2.0.1 changelog PR and release today.

parente:
https://pypi.org/project/jupyter_kernel_gateway/2.0.1/
https://github.com/conda-forge/jupyter_kernel_gateway-feedstock/pull/7

SimonBiggs:
Brilliant :). Thank you, thank you :)

PR #258, opened 2017-09-05
Don't check for auth token during OPTIONS preflight

SimonBiggs:
Details provided within https://github.com/jupyter/kernel_gateway/issues/257


kernel_gateway/mixins.py
```diff
@@ -65,7 +65,7 @@ def prepare(self):
         package.
         """
         server_token = self.settings.get('kg_auth_token')
-        if server_token:
+        if (server_token != '') & (self.request.method != 'OPTIONS'):
```
rolweber: That should be `and` instead of `&`. Python uses keywords for logical operators.
https://docs.python.org/2/library/stdtypes.html#boolean-operations-and-or-not
https://docs.python.org/2/library/stdtypes.html#bitwise-operations-on-integer-types

rolweber: Not sure if `!= ''` is the correct check here. Could server_token be null instead of the empty string?

parente: `if server_token and self.request.method != 'OPTIONS':`

parente: Also looks like a test is failing because `self.request` does not exist.

```
Traceback (most recent call last):
  File "/home/travis/build/jupyter/kernel_gateway/kernel_gateway/tests/test_mixins.py", line 49, in test_no_token_required
    self.mixin.prepare()
  File "/home/travis/build/jupyter/kernel_gateway/kernel_gateway/mixins.py", line 68, in prepare
    if (server_token != '') & (self.request.method != 'OPTIONS'):
AttributeError: 'TestableTokenAuthHandler' object has no attribute 'request'
```

parente:
@SimonBiggs thanks for submitting this. I'm a little miffed why `self.request` is not available in prepare after glancing at the tornado source for the RequestHandler base class. Maybe it's a test case problem.  Try fixing the if-statement per the comments above and let's see where we stand.

If you get stuck debugging the syntax and test problems, feel free to ping for help.

SimonBiggs:
Thanks for the feedback. In https://github.com/jupyter/kernel_gateway/blob/master/kernel_gateway/gatewayapp.py the following is given:

```python
@default('auth_token')
def _auth_token_default(self):
    return os.getenv(self.auth_token_env, '')
```

I think that makes it so that if auth_token isn't defined it will always be ''. I can explicitly test against '' or if I swap to using logical and instead of & I will also just be able to leave it as `if server_token`.

I'll submit a pull request soon.

SimonBiggs:
Is there a way to mock CORS requests within nose? I don't really know where to begin. Anyway, I'm just writing a test to cover this case with OPTIONS, however it would be nice to also test CORS  + authentication token directly.

kernel_gateway/tests/test_jupyter_websocket.py
```diff
@@ -150,6 +150,20 @@ def _set_api():
         self.assertRaises(ImportError, _set_api)
 
     @gen_test
+    def test_options_and_auth_token(self):
+        "OPTIONS requests should not need to submit a token."
```
parente: Make this triple-quoted so that it's a canonical docstring.

parente:
I haven't ever tried to mock a CORS request with Python. I think the test you've added to see if OPTIONS succeeds when a token is set provides enough test coverage and value.

SimonBiggs:
@parente Thanks for that.

kernel_gateway/mixins.py
```diff
@@ -64,8 +64,14 @@ def prepare(self):
         with the `@web.authenticated` decorated methods in the notebook
         package.
         """
+        request_is_options = False
+        if 'request' in dir(self):
```
rolweber: I'm not too firm in Python myself, but I think the calls to `dir()` are creating two temporary lists in this section, just to check whether there are attributes. Try the following instead (I haven't checked the exact syntax):
```python
   request_is_options = hasattr(self, 'request') and getattr(self.request, 'method', '') == 'OPTIONS'
```
https://stackoverflow.com/a/610893/5629418
https://stackoverflow.com/a/611708/5629418
This could also be inline below as
```python
if server_token and not (...):
```
But looking further below, `client_token` is also read from `self.request` without checking whether that exists, and a request without `method` wouldn't make any sense. I agree with @parente that it is probably a problem with the test setup if `self.request` doesn't exist.

kernel_gateway/tests/test_jupyter_websocket.py
```diff
@@ -150,6 +150,20 @@ def _set_api():
         self.assertRaises(ImportError, _set_api)
 
     @gen_test
+    def test_options_and_auth_token(self):
```
rolweber: Could you please rename this to `test_options_without_auth_token`?

kernel_gateway/tests/test_jupyter_websocket.py
```diff
@@ -150,6 +150,20 @@ def _set_api():
...
+        """OPTIONS requests should not need to submit a token."""
+        # Set token requirement
+        app = self.get_app()
```
rolweber: @parente: The docstring of [`get_app`](https://github.com/SimonBiggs/kernel_gateway/blob/03c53b4423f7fafccefbc259501405c60a438d83/kernel_gateway/tests/test_gatewayapp.py#L90) talks about a "Tornado test runner". Are we using that? If not, could that be the reason why `self.request` wasn't set? I haven't seen the original error message, I'm just second-guessing from the conversation.

rolweber:
I found the error message after all...
```
   AttributeError: 'TestableTokenAuthHandler' object has no attribute 'request'
```
That's not from a Tornado test, but from some mock I set up to test the auth handler. There is exactly one test which doesn't set the `request` attribute:
https://github.com/SimonBiggs/kernel_gateway/blob/master/kernel_gateway/tests/test_mixins.py#L46

@SimonBiggs, could you please just copy the `self.request = ` line from one of the other tests there, and get rid of the check for existence of `self.request` in the actual code?

rolweber:
@SimonBiggs: On second thought, could you add another test case for OPTIONS to the `test_mixins.py` file? The idea was to cover the complete logic of the auth mixin there.

SimonBiggs:
@rolweber @parente I think I managed to implement all of the feedback. Hope I didn't miss anything. This is a much nicer result.

rolweber:
@parente: I'll leave the final look and merge to you, in case I missed something.

parente:
LGTM to me. Thanks @SimonBiggs for the code and tests. 🍰  for your first contrib here.

SimonBiggs:
Thanks :). It's a pleasure :). Might this be able to be pushed out in a patch version release to pip?

PR #263, opened 2017-10-18
Fix #262: JSONErrorsMixin not returning correct results

### LINKED ISSUE #262: JSONErrorsMixin not returning correct results after Notebook 5.2.0 is installed
This is related to [Notebook Issue 2942](https://github.com/jupyter/notebook/issues/2942).

One of the changes in Notebook 5.2.0 was the removal of the `@json_errors` decorator via [PR 2853](https://github.com/jupyter/notebook/pull/2853).  Instead, much of the equivalent code was moved into the `APIHandler.write_error` method.  However, by virtue of the decorator's removal and the inheritance hierarchy, the `JSONErrorsMixin.write_error` method is called instead of `APIHandler.write_error`.  `JSONErrorsMixin.write_error` doesn't handle the traceback functionality present in `@json_errors` (or `APIHandler`) and results in the breakage of the `TestCustomDefaultKernel` test.

Since `APIHandler.write_error` should be doing the equivalent of `@json_errors`, I commented out the subclassing of `JSONErrorsMixin` from some of the handler classes, but then encountered a different set of test errors - these relating to no `reason` field on HTTP errors.  It turns out that one of the pieces of functionality in `@json_errors` not "transferred" to `APIHandler` was the appropriate setting of the reason code.

Since its not sufficient to fallback to using `APIHandler`, we should make the appropriate changes to `JSONErrorsMixin.write_error` in order to bridge the gap between Notebook 5.2 GA and the future release that addresses the APIHandler/reason issue.

kevin-bates:
After Notebook 5.2.0 is installed, JSONErrorsMixin.write_error is
used to convert errors into JSON responses.  Previously, Notebook's
`@json_errors` decorator performed this action. Since the decorator's
replacement, `APIHandler.write_error`, doesn't adequately set the
`reason` field, other tests fail so its not sufficient to drop the
use of JSONErrorsMixin.  As a result, JSONErrorsMixin now incorporates
the appropriate code from `APIHandler.write_error` (in particular the
setting of the `traceback` field) - in addition to appropriately
setting the `reason` field.  This mixin should continue to be used
in the handler class hierachy until `APIHandler` is fixed.

I have tested these changes with Notebook 5.1.0 and 5.2.0 installed.

Fixes #262.

### DIFF
--- a/kernel_gateway/mixins.py
+++ b/kernel_gateway/mixins.py
@@ -3,6 +3,8 @@
 """Mixins for Tornado handlers."""
 
 import json
+import traceback
+from tornado import web
 try:
     # py3
     from http.client import responses
@@ -85,12 +87,8 @@ class JSONErrorsMixin(object):
     def write_error(self, status_code, **kwargs):
         """Responds with an application/json error object.
 
-        Overrides the HTML renderer in the notebook server to force all errors
-        to JSON format. Outputs all errors as JSON in the same format as the
-        notebook.base.handlers.json_errors decorator. Avoids having to
-        (re-)decorate everything that the kernel gateway overrides. Also avoids
-        rendering errors as a human readable HTML pages in cases where the
-        decorator is not used in the notebook code base.
+        Overrides the APIHandler.write_error in the notebook server until it
+        properly sets the 'reason' field. 
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


rolweber:
I currently don't have time to dig deeper into the problem, but I had a look at the diff, and it looks good.

parente:
Thanks for digging into this issue. It's not the first time we've found we've found an incompatibility with new notebook versions because KG is definitely touches more than the public API of the notebook server.

Is notebook>=5.2.0 now a requirement after this patch goes in? Or will it work with notebook 5.0 and up? I'm asking to see if we need to (1) bump the required minimum notebook version to 5.2 now and (2) see if this should be a patch release or a minor release, assuming we're sticking with semantic versioning.

kevin-bates:
Hi @parente.  No, this update doesn't introduce any dependency change on notebook versions.  When notebook 5.1.0 (or less) is in play `@json_errors` builds the json exceptions, while with notebook 5.2.0, the `JSONErrorsMixin` builds the exception.  All nosetests pass with this change and either version of notebook.

In addition, once the notebook issue with `APIHandler.write_error` is fixed (to [set the `reason` field](https://github.com/jupyter/notebook/pull/2958)), then `JSONErrorsMixin` could be removed as a subclass on the various JKG handlers.  The tests that call the mixin directly could either be removed or the class could remain (not sure if its needed at that point).

parente:
Sorry for the delay on this one. I can work on backporting this to the 2.0.x branch to cut a 2.0.2 release with it soon.

rolweber:
@parente: Can you find some time for cutting a new release?

parente:
@rolweber @kevin-bates I've tagged 2.0.2 and pushed a release to pypi.

kevin-bates:
@parente - thank you!  Does something similar need to happen for anaconda envs? (Sorry, total newbie to this python stuff).  Since our tests run via anaconda envs it still looks like they're picking up 2.0.1 from conda-forge.
```
jupyter_kernel_gateway:             2.0.1-py_0         conda-forge
```

parente:
@kevin-bates Indeed. I'll cc on the PR over in the conda-forge feedstock.

kevin-bates:
Awesome. Thank you @parente!

PR #266, opened 2017-11-01
Additional run-time configurations for Kernel Gateway

### LINKED ISSUE #265: Make KernelSpecManager and KernelManager configurable
We want to make KernelSpecManager and kernel_manager_class configurable at run-time. This gives users flexibility to customize how Kernelspecs are fetched and how kernels are started.

shiti-saxena:
resolves #265

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -10,12 +10,14 @@ import nbformat
 import importlib
 import signal
 
+from notebook.services.kernels.kernelmanager import MappingKernelManager
+
 try:
     from urlparse import urlparse
 except ImportError:
     from urllib.parse import urlparse
 
-from traitlets import Unicode, Integer, default, observe
+from traitlets import Unicode, Integer, default, observe, Type, Instance
 
 from jupyter_core.application import JupyterApp, base_aliases
 from jupyter_client.kernelspec import KernelSpecManager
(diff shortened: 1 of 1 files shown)


kernel_gateway/gatewayapp.py
```diff
@@ -277,6 +279,27 @@ def keyfile_default(self):
...
+    kernel_spec_manager = Instance(KernelSpecManager, allow_none=True)
+
+    kernel_spec_manager_class = Type(
+        default_value=KernelSpecManager,
+        config=True,
+        help="""
+        The kernel spec manager class to use. Should be a subclass
+        of `jupyter_client.kernelspec.KernelSpecManager`.
+
+        The Api of KernelSpecManager is provisional and might change
```
rolweber: We shouldn't have API warnings from Jupyter Client in the kernel gateway code, I think.

shiti-saxena: I wasn't sure what the format for help would be so I copied it from [Jupyter Notebook](https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L997-L1007)

shiti-saxena:
@rolweber I've updated the PR with the suggested changes. Please review it.

parente:
The config options look reasonable to me for advanced users. Anyone looking to add a custom kernel manager will need to be careful to support all the mode and option behaviors of the default managers.

PR #272, opened 2017-12-29
fix a bug on getting the kernel language

gingerhot:
In the line 66, the `self.parent` here doesn't have an attribute named `seed_source` indeed. Since this line is wrapped by a `try`, we always get a `None`.  

In fact we should get the `language` from `self.parent.seed_notebook`,  guess maybe it's a typo of the former contributor?

### DIFF
--- a/kernel_gateway/notebook_http/__init__.py
+++ b/kernel_gateway/notebook_http/__init__.py
@@ -63,14 +63,17 @@ class NotebookHTTPPersonality(LoggingConfigurable):
         # Build the parser using the comment syntax for the notebook language
         func = getattr(cell_parser_module, 'create_parser')
         try:
-            lang = self.parent.seed_source['metadata']['language_info']['name']
+            kernel_language = self.parent.seed_notebook['metadata']['language_info']['name']
         except (AttributeError, KeyError):
-            lang = None
-        prefix = self.comment_prefix[lang]
+            kernel_language = None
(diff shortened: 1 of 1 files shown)


rolweber:
Oh, that explains the setting for `None` in the comment_prefix array.

kernel_gateway/notebook_http/__init__.py
```diff
@@ -63,14 +63,17 @@ def __init__(self, *args, **kwargs):
         # Build the parser using the comment syntax for the notebook language
         func = getattr(cell_parser_module, 'create_parser')
         try:
-            lang = self.parent.seed_source['metadata']['language_info']['name']
+            kernel_language = self.parent.seed_notebook['metadata']['language_info']['name']
         except (AttributeError, KeyError):
-            lang = None
-        prefix = self.comment_prefix[lang]
+            kernel_language = None
+        if kernel_language in self.comment_prefix:
```
rolweber: Instead of if/else, please use the pythonic lookup:

    kernel_language = self.comment_prefix.get(kernel_language, '#')

Then you can also remove the value for `None` in the comment_prefix dict.

gingerhot: ok~

rolweber:
Thanks for digging into this!

gingerhot:
My pleasure~

PR #275, opened 2018-02-02
fix link to http-mode.html

felixhummel:
Scratching my itch.
http://jupyter-kernel-gateway.readthedocs.io/en/latest/uses.html had a
bad link I clicked.

I should be doing this for the other links too, but I have to go to
sleep. ;)

This does not fix targets like PDF and man, but I guess HTML is the
primary use case anyways...


parente:
Thanks for the HTML fixes. 🍰

Issue #279, opened 2018-03-14
Meaning of env_whitelist and passing vars from os.environ.


GrahamDumpleton:
When using kernel_gateway on RHEL/CentOS with SCL (www.softwarecollections.org) versions of Python, there is a dependency on ``LD_LIBRARY_PATH`` being passed through to process created which use Python. If this isn't done, you end up with error such as:

```
[KernelGatewayApp] KernelRestarter: restarting kernel (3/5), new random ports
[KernelGatewayApp] Starting kernel: ['/opt/app-root/bin/python3', '-m', 'ipykernel_launcher', '-f', '/opt/app-root/src/.local/share/jupyter/runtime/kernel-5f5bb348-3c3a-4ec5-aea2-fafba2c0cf3a.json']
/opt/app-root/bin/python3: error while loading shared libraries: libpython3.5m.so.rh-python35-1.0: cannot open shared object file: No such file or directory
```

There does exist the configuration setting:

```
--JupyterWebsocketPersonality.env_whitelist=<List>
    Default: []
    Environment variables allowed to be set when a client requests a new kernel
```

but it doesn't have the desired affect of allowing you to white list additional environment variables from ``os.environ`` so they get passed. Instead, the white list is applied to the ``model['env']`` in code which am presuming is some description of the kernel being run.

* https://github.com/jupyter/kernel_gateway/blob/1bb6cb2d67493c16785d1aea20fc3543171a50e0/kernel_gateway/services/kernels/handlers.py#L57

```
            env.update({key: value for key, value in model['env'].items()
                   if key.startswith('KERNEL_') or key in self.env_whitelist})
```

What is ``model['env']`` in this situation? It comes through as empty.

Should the code perhaps be:

```
            env.update({key: value for key, value in os.environ.items()
                   if key.startswith('KERNEL_') or key in self.env_whitelist})
````

Or is there some other way of being able to white list environment variables that will be passed through to the kernel application process?

**Comment by GrahamDumpleton** [comment]
So this ``model['env']`` is environment variables passed across from the remote client triggering the creation of the kernel instance. It seems therefore need to look for a separate configuration setting which allows process environment variables from the process starting the kernel sub process to be passed through. Right now only ``PATH`` is passed through.

```
            # Start with the PATH from the current env. Do not provide the entire environment
            # which might contain server secrets that should not be passed to kernels.
            env = {'PATH': os.getenv('PATH', '')}
```

So add after this:

```
[...]

PR #280, opened 2018-03-14
Add whitelist for environment variables to be inherited from gateway process by kernel.

GrahamDumpleton:
As raised in #279, this PR adds ability to define a whitelist of environment variables which are allowed to be inherited from the kernel gateway process, when spawning a kernel process. This is needed to allow environment variables such as ``LD_LIBRARY_PATH``, ``LD_PRELOAD`` or other special environment variables to be passed through when executing the kernel process. In the case of ``LD_LIBRARY_PATH`` this avoids error such as:

```
[KernelGatewayApp] KernelRestarter: restarting kernel (3/5), new random ports
[KernelGatewayApp] Starting kernel: ['/opt/app-root/bin/python3', '-m', 'ipykernel_launcher', '-f', '/opt/app-root/src/.local/share/jupyter/runtime/kernel-5f5bb348-3c3a-4ec5-aea2-fafba2c0cf3a.json']
/opt/app-root/bin/python3: error while loading shared libraries: libpython3.5m.so.rh-python35-1.0: cannot open shared object file: No such file or directory
```

The list of environment variables allowed to be inherited can be set by ``KG_ENV_PROCESS_WHITELIST`` or ``KernelGatewayApp.env_process_whitelist``. Used ``env_process_whitelist`` instead of ``process_whitelist`` on ``KernelGatewayApp`` in the end so clearly distinguished from ``JupyterWebsocketPersonality.env_whitelist`` which is whitelist of environment variables that can be passed from client making request to start kernel. Need to be different to also avoid conflict with ``KG_ENV_WHITELIST`` environment variable.

### DIFF
--- a/docs/source/config-options.md
+++ b/docs/source/config-options.md
@@ -179,6 +179,10 @@ KernelGatewayApp options
     Default: None
     Runs the notebook (.ipynb) at the given URI on every kernel launched. No
     seed by default. (KG_SEED_URI env var)
+--KernelGatewayApp.env_process_whitelist=<List>
+    Default: []
+    Environment variables allowed to be inherited from current process by a
+    new kernel.
 
 NotebookHTTPPersonality options
 -------------------------------
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -17,7 +17,7 @@ try:
 except ImportError:
     from urllib.parse import urlparse
 
-from traitlets import Unicode, Integer, default, observe, Type, Instance
+from traitlets import Unicode, Integer, default, observe, Type, Instance, List
 
 from jupyter_core.application import JupyterApp, base_aliases
 from jupyter_client.kernelspec import KernelSpecManager
@@ -232,6 +232,14 @@ class KernelGatewayApp(JupyterApp):
     def force_kernel_name_default(self):
         return os.getenv(self.force_kernel_name_env, '')
 
+    env_process_whitelist_env = 'KG_ENV_PROCESS_WHITELIST'
+    env_process_whitelist = List(config=True,
+                                 help="""Environment variables allowed to be inherited from the spawning process by the kernel""")
+
+    @default('env_process_whitelist')
+    def env_process_whitelist_default(self):
+        return os.getenv(self.env_process_whitelist_env, '').split(',')
+
     api_env = 'KG_API'
     api_default_value = 'kernel_gateway.jupyter_websocket'
     api = Unicode(api_default_value,
@@ -444,6 +452,7 @@ class KernelGatewayApp(JupyterApp):
             kg_expose_headers=self.expose_headers,
             kg_max_age=self.max_age,
             kg_max_kernels=self.max_kernels,
+            kg_env_process_whitelist=self.env_process_whitelist,
             kg_api=self.api,
             kg_personality=self.personality,
             # Also set the allow_origin setting used by notebook so that the
--- a/kernel_gateway/services/kernels/handlers.py
+++ b/kernel_gateway/services/kernels/handlers.py
@@ -23,6 +23,10 @@ class MainKernelHandler(TokenAuthorizationMixin,
     def env_whitelist(self):
         return self.settings['kg_personality'].env_whitelist
 
+    @property
+    def env_process_whitelist(self):
+        return self.settings['kg_env_process_whitelist']
+
     @gen.coroutine
     def post(self):
         """Overrides the super class method to honor the max number of allowed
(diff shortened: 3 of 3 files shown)


rolweber:
Travis CI detects problems with all tested Python versions except 3.4.
I'm not sure if this is something in the PR, or an infrastructure problem.

rolweber:
The code changes look good to me.

Unfortunately, I currently don't have the time to dig further into the Travis failures. @parente, would you have some cycles to spare?

rolweber:
This is the Travis CI problem for all failing versions:
```txt
ERROR: Failure: ImportError (cannot import name LogTrapTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/build/jupyter/kernel_gateway/kernel_gateway/tests/test_gatewayapp.py", line 10, in <module>
    from tornado.testing import AsyncHTTPTestCase, LogTrapTestCase
ImportError: cannot import name LogTrapTestCase
```
The PR doesn't mention `LogTrapTestCase` anywhere, so the problem is not in the code changes. Since this is Tornado related, it might have to do with the release of Tornado 5.0. I've seen that cause problems elsewhere, when the latest version is installed automatically.

rolweber:
For example https://github.com/jupyter/notebook/issues/3023

rolweber:
Indeed, Tornado 5.0 dropped support for Python 3.3. So our Travis CI tests are failing on all Python versions for which Tornado 5.0 is available.

kevin-bates:
@rolweber - We ran into this [same problem in Enterprise Gateway](https://github.com/jupyter-incubator/enterprise_gateway/pull/277) (since all unittests are derived from JKG) and I had to block use of tornado >= 5.0.  `LogTrapTestCase` was deprecated long ago and now removed in 5.0.  However, its removal or replacement with the suggested class `ExpectLog` isn't straightforward and I could take the time to figure it out.  JEG and presumably JKG work fine otherwise with tornado 5.0, but I wanted to clean solution, then was going to apply the same here.

kevin-bates:
@GrahamDumpleton @rolweber - Regarding the code changes for this PR, isn't the `env:` stanza in the kernelspec (i.e., `kernel.json`) file how things like server-side envs are addressed?  For example, here's such an [env stanza](https://github.com/jupyter-incubator/enterprise_gateway/blob/master/etc/kernelspecs/spark_python_yarn_cluster/kernel.json#L7-L13) from one of the kernelspecs (samples) we provide in Enterprise Gateway.

I'm 99% certain that the `env:` stanza is what seeds the `model` structure that you referenced in #279.  So if the `LD_` envs you need where configured into the `kernel.json` file, you'd be good to go.  In addition, some envs may not apply to all kernels (especially language-specific ones) and since JKG (and JEG) are intended to serve up multitudes of kernels, I think its best to let the existing configuration strategies take their course.

As a result, I'd have to down-vote this PR unless I'm missing something.

rolweber:
kernelspecs are one way to define env variables. But if the setting is not specific to the kernel, but rather to the Jupyter installation, it makes sense to pass environment variables from the Jupyter process to the kernels as well. Then you can use kernelspecs you find somewhere on the internet unmodified.

GrahamDumpleton:
Yep, Roland got it. This isn't trying to set environment variables because a specific kernel needed them, but that they were needed for underlying stuff such as Python itself, or some other reason related to the runtime.

For example, in container image I am running stuff in, I also need to ensure that ``LD_PRELOAD``, ``NSS_WRAPPER_PASSWD`` and ``NS_WRAPPER_GROUP`` are passed through. These  are used to force load a shared library into all applications which fakes out passwd/group entries for users that don't exist in system passwd and group files. If don't return valid entries, various Python packages can fail. In Jupyter ``docker-stacks`` images didn't need that NSS mess luckily as used better solution of making passwd and group files writable so could add entries from ``start.sh``.

* https://github.com/jupyter/docker-stacks/blob/master/base-notebook/start.sh#L70

As to having ``PATH`` in ``env_process_whitelist`` I saw that as too dangerous as you have to have that. If a user overrides the configuration and doesn't include ``PATH`` in their overrides, it would break things. So ``PATH`` has to always be there no matter what.

lresende:
LGTM, the CI build issues are unrelated and I will submit a fix for that.

rolweber:
I merged the fix for the Travis CI builds and requested a new build for this PR.

PR #282, opened 2018-04-04
Limit tornado version to avoid build issues

lresende:
Tornado version 5.0 was recently released and introduces significant
changes. One change that breaks Travis builds is the removal of the
tornado.testing.LogTrapTestCase. Attempts to remove that reference
still lead to failures. This change mostly limit the tornado version
to temporarily avoid build issues.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,5 +2,5 @@ jupyter_core>=4.0
 jupyter_client>=4.2.0
 notebook>=5.0.0
 traitlets>=4.2.0
-tornado>=4.2.0
+tornado>=4.2.0,<5.0
 requests>=2.7,<3.0
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ containerized and scaled out using common technologies like
         'jupyter_client>=4.2.0',
         'notebook>=5.0.0,<6.0',
         'traitlets>=4.2.0',
-        'tornado>=4.2.0',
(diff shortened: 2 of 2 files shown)


rolweber:
Hmm... Travis doesn't care. Here's what's being executed by Travis:
https://github.com/jupyter/kernel_gateway/blob/master/.travis.yml

I'm not familiar with all of the commands, but from the logs, it seems that
```bash
python setup.py bdist_wheel sdist
```
Builds an installable package from the KG sources, and the following line then installs that, including the dependencies. That's where Tornado 5 is being picked up:
https://travis-ci.org/jupyter/kernel_gateway/jobs/361960749#L680

The `requirements.txt` file doesn't come into play at all. I see two cheap options to fix the test runs:
- Change `setup.py` itself.
- Add another line to `.travis.yml` that downgrades Tornado. That would go just before the `pip freeze`.

@lresende: Do you want to give it another try?

lresende:
I have missed the setup.py, thanks for catching that @rolweber. And I believe we don't need to change travis config for this. With the updated commit the build is back to green again.

rolweber:
The difference is that now, we have the dependency on Tornado < 5 in the packages built from master. If we change .travis.yml instead of setup.py, only the test runs would be affected. But as we have to sort this out before a new release anyway, I'm fine with this change.

PR #284, opened 2018-04-23
Add reason argument to set_status() so that custom messages flow

### LINKED ISSUE #283: Add ability to return custom error messages.
This is related to [Enterprise Gateway Issue 296](https://github.com/jupyter-incubator/enterprise_gateway/issues/296).  With the current implementation of `JSONErrorsMixin.write_error()` it is not possible to do the following:
1. Return custom error text associated with a status code.
2. Return custom status codes.

By extending the `set_status()` method with a `reason` named argument, both of the above can happen.

kevin-bates:
Without adding a named argument of `reason` to tornado's `set_status()`
method, messages associated with a thrown HTTPError will not be returned
to the caller.  Instead, the reason corresponding to the built-in status
code is returned.  In addition, because the reason argument is used,
enforcement of a built in status code is skipped, thereby allowing for
the optional use of non-builtin status codes.

Fixes #283

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,5 +56,8 @@ docs/_build/
 # PyBuilder
 target/
 
+# PyCharm
+.idea/
+
 .DS_Store
 .ipynb_checkpoints/
--- a/kernel_gateway/mixins.py
+++ b/kernel_gateway/mixins.py
@@ -125,5 +125,5 @@ class JSONErrorsMixin(object):
                 reply['reason'] = custom_reason
 
         self.set_header('Content-Type', 'application/json')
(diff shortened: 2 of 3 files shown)


kevin-bates:
Hmm - looks like this breaks the tests.  Need to determine what's going on here.

kevin-bates:
I've amended the PR such that `TestableJSONErrorsHandler` has a compatible `set_status()` signature to that of Tornado's.

parente:
PR seems reasonable to me. (Pun intended.) Thanks @kevin-bates!

PR #292, opened 2018-07-05
Update dependencies for Jupyter components.

### LINKED ISSUE #291: Update dependencies for jupyter framework
The dependencies upon lower level jupyter components have gotten to a point that they can be problematic for those installing kernel gateway on top of older versions.  [This issue](https://github.com/jupyter/kernel_gateway_demos/issues/61) can occur when older versions of jupyter_core (`< 4.4.0`) are present while updates to jupyter_client have occurred.  

Since both Notebook (below) and Enterprise Gateway (above) are using updated dependencies, we should probably do the same for Kernel Gateway.  The updates would include jupyter_core(`>=4.4.0`), jupyter_client(`>=5.2.0`) and notebook (`>=5.3.0,<6.0`)

kevin-bates:
Bring the dependencies for jupyter components (namely jupyter_core) to
be more current.

Fixes #291

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
-jupyter_core>=4.0
-jupyter_client>=4.2.0
-notebook>=5.0.0
+jupyter_core>=4.4.0
+jupyter_client>=5.2.0
+notebook>=5.3.0,<6.0
 traitlets>=4.2.0
 tornado>=4.2.0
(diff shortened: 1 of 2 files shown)


parente:
Thanks @kevin-bates. Since we're updating dependency requirements, we should make sure the next release of KG includes a minor version bump, not just a patch release.

PR #294, opened 2018-07-17
Allow access from remote hosts

### LINKED ISSUE #293: Allow access from remote hosts
As of [Notebook PR 3714](https://github.com/jupyter/notebook/pull/3714/) access from non-local hosts is being denied by default.  Since this is the primary functionality provided by Kernel Gateway, we should always enable the ability to provide access from remote hosts.  (The [same change](https://github.com/jupyter-incubator/enterprise_gateway/pull/386) was required in Enterprise Gateway.)

kevin-bates:
A recent change to Notebook (5.6.0) prevents access from remote hosts by default.
This change unconditionally re-enables that ability since remote access is a primary
tenet of Kernel Gateway.

Fixes #293


bashterm:
Was this an issue in release 2.0.2? KG prevents me from accessing from a remote host, even when i set NotebookApp.allow_remote_access to true.

kevin-bates:
This is purely a function of the underlying Notebook release.  What is the nature of the your issues?

bashterm:
I'm running JKG in a container with port 8888 forwarded through to my localhost. When I try and connect to it from my client using an HTTP GET request for /api/kernelspecs it returns a 403 forbidden. 

The log says something like Access from remote host forbidden. Set NotebookApp.allow_remote_access to allow access.

The kernelgateway variable allow_origins is set to '*'.

parente:
Until there's a release of KG with this change, the problem will continue to occur for users who install KG since the notebook dependency is listed as `notebook>=5.3.0,<6.0`.

kevin-bates:
I'm sorry @bashterm - I must have had a complete brain freeze.  I was under the impression you were already running the JKG tip.  So, if you're not on the tip *and* are running Notebook >= 5.6.0, then yes, @parente is correct, you need the latest JKG (and we should build a release).  Especially since NB 5.6.0 was pushed out to address a security issue.  Of course, your workaround is to downgrade NB < 5.6.0, but that's probably a function of the kind of environment you're dealing with.

Regarding your question about Enterprise Gateway, yes, our [0.9.4 release](https://github.com/jupyter-incubator/enterprise_gateway/releases/tag/v0.9.4) has the fix.  You should be able to install and run like JKG - then you can begin exploring the remote kernel capabilities.  We hope to have our 1.0 release by [JupyterCon](https://conferences.oreilly.com/jupyter/jup-ny).

bashterm:
Ok. I have a lot of questions about Enterprise gateway. Should I just open an issue regarding them over on the GitHub for the Enterprise Gateway project?

kevin-bates:
Yes, please.

PR #299, opened 2018-08-08
adds x_header configuration option for use behind proxies

vlaurenzano:
This passes the xheaders configuration option to tornado. The option informs tornado to use the forwarded X-Real-Ip header if given. Useful for when the kernelgateway is behind a proxy and you need the original ip (for logging for instance).

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -18,7 +18,7 @@ try:
 except ImportError:
     from urllib.parse import urlparse
 
-from traitlets import Unicode, Integer, default, observe, Type, Instance, List
+from traitlets import Unicode, Integer, default, observe, Type, Instance, List, CBool
 
 from jupyter_core.application import JupyterApp, base_aliases
 from jupyter_client.kernelspec import KernelSpecManager
@@ -176,12 +176,12 @@ class KernelGatewayApp(JupyterApp):
         return os.getenv(self.expose_headers_env, '')
 
     trust_xheaders_env = 'KG_TRUST_XHEADERS'
-    trust_xheaders = Unicode(config=True,
+    trust_xheaders = CBool(False, config=True,
         help='Use x-* header values for overriding the remote-ip, useful when application is behing a proxy. (KG_TRUST_XHEADERS env var)'
     )
     @default('trust_xheaders')
     def trust_xheaders_default(self):
-        return os.getenv(self.trust_xheaders_env, 'False')
+        return strtobool(os.getenv(self.trust_xheaders_env, 'False'))
 
 
     max_age_env = 'KG_MAX_AGE'
@@ -515,7 +515,7 @@ class KernelGatewayApp(JupyterApp):
         """
         ssl_options = self._build_ssl_options()
         self.http_server = httpserver.HTTPServer(self.web_app,
-                                                 xheaders=strtobool(self.trust_xheaders),
+                                                 xheaders=self.trust_xheaders,
                                                  ssl_options=ssl_options)
 
         for port in random_ports(self.port, self.port_retries+1):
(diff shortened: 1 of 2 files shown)


vlaurenzano:
I set it up as to not change current behavior (through a feature flag) but it probably makes sense to set it to true by default and maybe even forget the flag.

rolweber:
This looks like a very specific solution. I wonder if we could introduce something more generic instead. Notebook has a `tornado_settings` [config option](https://jupyter-notebook.readthedocs.io/en/stable/config.html). Would that work for your xheaders configuration, too?

vlaurenzano:
Looking at Notebook now on your suggestion and it looks like they already have the xheaders flag. The tornado_settings option won't work because it's used for the Application and not the httpServer.

It seems notebook actually has an xheaders config:
https://github.com/jupyter/notebook/blob/7674331e3d10881ea6727e7c6c8daff78f405756/notebook/notebookapp.py#L1375

It looks like the tornado settings are used for the app and not the server:
https://github.com/jupyter/notebook/blob/7674331e3d10881ea6727e7c6c8daff78f405756/notebook/notebookapp.py#L1353

rolweber:
I see. Thanks for checking :-)

I'd appreciate if someone else also has a look at this before merging.

kernel_gateway/gatewayapp.py
```diff
@@ -174,6 +174,16 @@ def allow_origin_default(self):
     def expose_headers_default(self):
         return os.getenv(self.expose_headers_env, '')
 
+    x_headers_env = 'KG_X_HEADERS'
+    x_headers = Unicode(config=True,
```
kevin-bates: Since it sounds like this is providing the identical functionality to that provided in Notebook, then I suggest we use the same option name (and variable): `trust_xheaders`.  This fits the boolean datatype better as well, since `x_headers` sounds more like a noun.

I also suggest using the `Bool` traitlet rather than `Unicode`.  This should eliminate the need for `strtobool` and import.

kernel_gateway/gatewayapp.py
```diff
@@ -174,6 +174,16 @@ def allow_origin_default(self):
...
+        help='Uses x-* header values for overriding the remote-ip, useful when application is behing a proxy. (KG_X_HEADERS env var)'
+    )
+
+    @default('x_headers')
+    def x_headers_default(self):
+        return os.getenv(self.x_headers_env, 'false')
```
kevin-bates: It may be a good idea to handle various formats for the env value, so something like this may work more reliably...
```
return bool(os.getenv(self.x_headers_env, 'false').lower() == 'true')
```

kernel_gateway/gatewayapp.py
```diff
@@ -505,6 +515,7 @@ def init_http_server(self):
         """
         ssl_options = self._build_ssl_options()
         self.http_server = httpserver.HTTPServer(self.web_app,
+                                                 xheaders=bool(strtobool(self.x_headers)),
```
kevin-bates: Per above, I think we can eliminate the casts (and import) by using the traitlets - as Notebook did.

kernel_gateway/gatewayapp.py
vlaurenzano: The reason I used unicode and not Bool is because it is coming from an env variable and it didn't say in the documentation whether it would cast from a string or how. There is also a CBool, for casting bool, however it seems like it only works for integers. I'll look into it and update it with Bool or CBool if appropriate.

Agree about the name, I didn't know it existed in notebook prior to implementing.

kevin-bates: Sounds good.  Once the updates are in, we'll merge.  Thanks.

kevin-bates:
Squashed and merged - thanks again!

PR #304, opened 2018-09-17
Add ability for Kernel Gateway to ignore SIGHUP signal

tf3193:
We noticed that when having kernel gateway run for extended periods of time the process would terminate when the user would logoff or if we killed the terminal that started the Kernel Gateway. We found out that python 3.x will not ignore the SIGHUP signal even when ran with nohup, this lead us to this solution.

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -139,6 +139,15 @@ class KernelGatewayApp(JupyterApp):
     def allow_credentials_default(self):
         return os.getenv(self.allow_credentials_env, '')
 
+    allow_persistence_env = 'KG_ALLOW_PERSISTENCE'
+    allow_persistence = CBool(False, config=True,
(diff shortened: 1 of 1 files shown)


tf3193:
Thanks for the reply Kevin. I see no issue with making it a default behavior, I was just trying to preserve default behavior and be non intrusive to current users. If you would like I can remove the config hooks.

kevin-bates:
Awesome - yeah, let's just make it unconditional.  I'll wait until tomorrow to merge for others to comment. 

Would you mind applying this to the [Enterprise Gateway project](https://github.com/jupyter/enterprise_gateway/blob/master/enterprise_gateway/enterprisegatewayapp.py#L268-L269)?  You'll find it familiar. :smile:

tf3193:
All done Kevin! 
Sure I will open up a pull request on Enterprise Gateway as well! 

Thank you

kevin-bates:
@tf3193 - thank you for the contribution.
@rolweber - thank you for your review.

PR #314, opened 2019-02-22
Support serving kernelspec resources

kevin-bates:
A long-standing TODO item for NB2KG was to add support for kernelspec resources (`kernel.js`, `kernel.css` and `logo-*`).  This can't be accomplished until JKG provides the endpoint such that specific resources can be fetched.  This change hooks up the Notebook's kernelspecs resource handler - after which requests specifying the kernelspec name and file within will be returned:
`http://127.0.0.1:8889/kernelspecs/python3/logo-64x64.png`.

### DIFF
--- a/kernel_gateway/services/kernelspecs/handlers.py
+++ b/kernel_gateway/services/kernelspecs/handlers.py
@@ -3,12 +3,19 @@
 """Tornado handlers for kernel specs."""
 
 import notebook.services.kernelspecs.handlers as notebook_handlers
+import notebook.kernelspecs.handlers as notebook_kernelspecs_resources_handlers
+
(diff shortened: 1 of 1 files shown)


rolweber:
Just wondering: should we expect updates to the changelog in the same PR that adds an enhancement? Or are we going to collect the changelog updates when preparing for a new release?

kevin-bates:
Thank you for your review and merge Roland.  

I haven't really worked with change logs, but that makes sense to include updates in the associated PRs.  Since that is water under the bridge, I suppose it's best to queue this up until we cut the release.

Speaking of which... would you object to me cutting 2.2.0 so the NB2KG clients can leverage resources once that PR is merged?

rolweber:
I certainly would not object :-)

PR #315, opened 2019-03-15
Call tornado StaticFileHandler.get() as a coroutine

lresende:


lresende:
These changes support Tornado version 6.0 and above, and help fix the Enterprise Gateway build issues while using Tornado 6.0.1

PR #316, opened 2019-03-30
Remove code health landscape.io badge

willingc:
It no longer seems to be actively working.


rolweber:
It still seems to be working, but links with the branch name fail. This one shows current reports:
https://landscape.io/github/jupyter/kernel_gateway
On that page, there are links which in turn lead to an error page.

We could simply remove the branch name `/master` from the link address.

rolweber:
Thank you for the PR!

PR #321, opened 2019-08-02
Allow notebook==6.* since it seems to work for me.

juliusvonkohout:


lresende:
But we should also have a clean build before we merge it

juliusvonkohout:
Thanks @lresende . Will there be a new pypi release?

PR #324, opened 2019-08-11
Release 2.4.0

lresende:


kernel_gateway/_version.py
```diff
@@ -5,7 +5,7 @@
 
 version_info = (
     2,
-    4,
+    5,
```
rolweber: Shouldn't this version remain on 2.4.0 until the release has been tagged?

rolweber: Never mind... the 'dev' below needs to be removed for the release build anyway, so the version number can be fixed on the same occasion.

kevin-bates:
I suspect Luciano forgot to push the tag relative to https://github.com/jupyter/kernel_gateway/pull/324/commits/ba9296fefd80dcdcc6560bd8598f2c18cab4cc68

rolweber:
Ah, I see the tag now: https://github.com/jupyter/kernel_gateway/releases
I probably just missed that yesterday, looking only at the PR.

PR #338, opened 2020-08-10
Use appropriate maybe-future to handle asyncio futures

### LINKED ISSUE #337: POST api/kernel  return 'Kernel does not exist'
Hi,

I installed kernel gateway via conda.

When I try to create a new kernel using POST I get:

[W 200806 14:40:23 web:1786] 404 POST /api/kernels (xxx.xxx.xxx.xxx): Kernel does not exist: <coroutine object MappingKernelManager.start_kernel at 0x7f2990577680>

I have tried using the --debug flag but I had not found any other error message.

Any hint to troubleshoot this?

PD: The kernelspec endpoint returns the kernels that already exist in the jupyter notebook installation.

kevin-bates:
The method to handle conditional returns of futures in tornado was both deprecated and incompatible with asyncio.  This incompatibility surfaces with Notebook 6.1 since it uses asyncio. Now using `maybe_future` from `notebook.utils` instead.

Updated the python matrix in Travis.

Fixes #337

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,10 @@ dist: xenial   # required for Python >= 3.7
 sudo: false
 language: python
 python:
-    - 2.7
-    - 3.4
     - 3.5
     - 3.6
     - 3.7
+    - 3.8
 
 env:
     - ASYNC_TEST_TIMEOUT=10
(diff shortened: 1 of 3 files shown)


kevin-bates:
Thanks for the review @lresende.

> we should really look into transitioning users to EG which has been merged with JKG and supports both local and remote kernels

True, but EG doesn't support the http personality: `--KernelGatewayApp.api='kernel_gateway.notebook_http'` so JKG is still necessary for that.

PR #340, opened 2020-08-13
enable ssl_version as a JKG config option

### LINKED ISSUE #339: Unable to connect to jupyter kernel gateway on Safari Version >= 13.1
Hey,

I am unable to connect to the jupyter kernel over the web socket on Safari. 
I get this error: `WebSocket network error: OSStatus error - 9801: Cipher Suite negotiation failure`

Any idea how this can be fixed? 

PS: Big fan of the project ✌️

applecool:
Fixes #339

### DIFF
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -308,7 +308,7 @@ class KernelGatewayApp(JupyterApp):
     @default('ssl_version')
     def ssl_version_default(self):
         ssl_from_env = os.getenv(self.ssl_version_env)
-        return self.ssl_version_default_value if ssl_from_env is None else int(ssl_from_env)
+        return ssl_from_env if ssl_from_env is None else int(ssl_from_env)
 
     kernel_spec_manager = Instance(KernelSpecManager, allow_none=True)
 
@@ -506,7 +506,7 @@ class KernelGatewayApp(JupyterApp):
             ssl_options['keyfile'] = self.keyfile
         if self.client_ca:
             ssl_options['ca_certs'] = self.client_ca
-        if self.certfile and self.ssl_version:
+        if self.ssl_version:
             ssl_options['ssl_version'] = self.ssl_version
         if not ssl_options:
             # None indicates no SSL config
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -9,6 +9,7 @@ import os
 import sys
 import signal
 import socket
+import ssl
 from distutils.util import strtobool
 
 import nbformat
@@ -52,7 +53,8 @@ aliases.update({
     'seed_uri': 'KernelGatewayApp.seed_uri',
     'keyfile': 'KernelGatewayApp.keyfile',
     'certfile': 'KernelGatewayApp.certfile',
-    'client-ca': 'KernelGatewayApp.client_ca'
+    'client-ca': 'KernelGatewayApp.client_ca',
+    'ssl_version': 'KernelGatewayApp.ssl_version'
 })
 
 
@@ -298,6 +300,15 @@ class KernelGatewayApp(JupyterApp):
     def client_ca_default(self):
         return os.getenv(self.client_ca_env)
 
+    ssl_version_env = 'KG_SSL_VERSION'
+    ssl_version_default_value = ssl.PROTOCOL_TLSv1_2
+    ssl_version = Integer(ssl_version_default_value, config=True, allow_none=True,
+                        help="""Sets the SSL version to use for the web socket connection. (KG_SSL_VERSION env var)""")
+    
+    @default('ssl_version')
+    def ssl_version_default(self):
+        return int(os.getenv(self.ssl_version_env, self.ssl_version_default_value))
+
     kernel_spec_manager = Instance(KernelSpecManager, allow_none=True)
 
     kernel_spec_manager_class = Type(
@@ -494,12 +505,12 @@ class KernelGatewayApp(JupyterApp):
             ssl_options['keyfile'] = self.keyfile
         if self.client_ca:
             ssl_options['ca_certs'] = self.client_ca
+        if self.ssl_version:
+            ssl_options['ssl_version'] = self.ssl_version
         if not ssl_options:
             # None indicates no SSL config
             ssl_options = None
         else:
-            # SSL may be missing, so only import it if it's to be used
-            import ssl
             # Disable SSLv3 by default, since its use is discouraged.
             ssl_options.setdefault('ssl_version', ssl.PROTOCOL_TLSv1)
             if ssl_options.get('ca_certs', False):
--- a/kernel_gateway/gatewayapp.py
+++ b/kernel_gateway/gatewayapp.py
@@ -302,12 +302,13 @@ class KernelGatewayApp(JupyterApp):
 
     ssl_version_env = 'KG_SSL_VERSION'
     ssl_version_default_value = ssl.PROTOCOL_TLSv1_2
-    ssl_version = Integer(ssl_version_default_value, config=True, allow_none=True,
+    ssl_version = Integer(None, config=True, allow_none=True,
                         help="""Sets the SSL version to use for the web socket connection. (KG_SSL_VERSION env var)""")
     
     @default('ssl_version')
     def ssl_version_default(self):
-        return int(os.getenv(self.ssl_version_env, self.ssl_version_default_value))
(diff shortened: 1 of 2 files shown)


kernel_gateway/gatewayapp.py
```diff
@@ -298,6 +300,15 @@ def keyfile_default(self):
...
         return os.getenv(self.client_ca_env)
 
+    ssl_version_env = 'KG_SSL_VERSION'
+    ssl_version_default_value = ssl.PROTOCOL_TLSv1_2
+    ssl_version = Integer(ssl_version_default_value, config=True, allow_none=True,
+                        help="""Sets the SSL version to use for the web socket connection. (KG_SSL_VERSION env var)""")
+    
+    @default('ssl_version')
+    def ssl_version_default(self):
+        return int(os.getenv(self.ssl_version_env, self.ssl_version_default_value))
```
kevin-bates, line 310: I think we might need to follow suit with how the cert-related attributes are handled - which essentially default to None.  This means you'll need to do a little extra processing in `ssl_version_default()` to convert the numeric string (if an ENV is present) to an int, otherwise return None.

kernel_gateway/gatewayapp.py
```diff
@@ -494,12 +505,12 @@ def _build_ssl_options(self):
...
+        if self.ssl_version:
+            ssl_options['ssl_version'] = self.ssl_version
         if not ssl_options:
             # None indicates no SSL config
             ssl_options = None
         else:
-            # SSL may be missing, so only import it if it's to be used
-            import ssl
             # Disable SSLv3 by default, since its use is discouraged.
             ssl_options.setdefault('ssl_version', ssl.PROTOCOL_TLSv1)
```
kevin-bates, line 515: You could then update this to reference `self.ssl_version_default_value`.  We should probably delete the `# Disable` comment as well.

I'm not sure how critical it was to defer the import of SSL, but it _could_ just be referenced here and forgo its setup on `self.ssl_version_default_value` in that case.  But it appears to be part of the standard Python library, so it doesn't seem like a big deal to defer its import.  (I'm still relative new to Python, so I might be missing something about this.)

kernel_gateway/tests/test_gatewayapp.py
```diff
@@ -43,6 +43,7 @@ def test_config_env_vars(self):
         os.environ['KG_KEYFILE'] = '/test/fake.key'
         os.environ['KG_CERTFILE'] = '/test/fake.crt'
         os.environ['KG_CLIENT_CA'] = '/test/fake_ca.crt'
+        os.environ['KG_SSL_VERSION'] = '5'
```
kevin-bates, line 46: Since the new default (`PROTOCOL_TLSv1_2`) has a value of 5 anyway, I'm not sure this test is ensuring the code is doing the right thing.  I would recommend using '3' and ensuring that.

kernel_gateway/tests/test_gatewayapp.py
```diff
@@ -67,6 +68,7 @@ def test_config_env_vars(self):
         self.assertEqual(app.keyfile, '/test/fake.key')
         self.assertEqual(app.certfile, '/test/fake.crt')
         self.assertEqual(app.client_ca, '/test/fake_ca.crt')
+        self.assertEqual(app.ssl_version, 5)
         self.assertEqual(app.trust_xheaders, False)
 
     def test_trust_xheaders(self):
```
kevin-bates, line 74: Since the defaulting of `ssl_options` will be deferred, it might be helpful to have a test that ensures the default of `5` is being used when no options other than one of the cert options are specified.

applecool:
@kevin-bates Please take a look :)

kernel_gateway/gatewayapp.py
```diff
@@ -494,14 +506,13 @@ def _build_ssl_options(self):
             ssl_options['keyfile'] = self.keyfile
         if self.client_ca:
             ssl_options['ca_certs'] = self.client_ca
+        if self.certfile and self.ssl_version:
```
kevin-bates, line 509: I think we should try to keep ssl_version independent from the related items for determining whether it should be set.

kernel_gateway/tests/test_gatewayapp.py
```diff
@@ -77,7 +79,11 @@ def test_trust_xheaders(self):
         app = KernelGatewayApp()
         self.assertEqual(app.trust_xheaders, True)
 
-
+    def test_ssl_options(self):
+        app = KernelGatewayApp()
+        os.environ['KG_CERTFILE'] = '/test/fake.crt'
```
kevin-bates, line 84: We should probably forgo setting the cert file here (per previous comment).

kernel_gateway/gatewayapp.py
applecool: But `certfile` key is required on the ssl_options according to the comment I found in the `__init__` function in tornado's `tcpserver.py`. The comment is: `# Only certfile is required: it can contain both keys`. That means, we can't really set `ssl_version` attribute just by itself. If I do that I would get the same `KeyError('missing key "certfile" in ssl_options')`. Right?

kevin-bates: Thanks for the reference.  While that comment is true, dependent applications shouldn't have to understand the inter-relationships.  You'll get that `KeyError` today if you only set `keyfile` or `ca_certs` as well, and we don't enforce that relationship here or in the tests.

applecool: Got it. I will remove that dependency.

kernel_gateway/gatewayapp.py
```diff
@@ -308,7 +308,7 @@ def client_ca_default(self):
     @default('ssl_version')
     def ssl_version_default(self):
         ssl_from_env = os.getenv(self.ssl_version_env)
-        return self.ssl_version_default_value if ssl_from_env is None else int(ssl_from_env)
+        return ssl_from_env if ssl_from_env is None else int(ssl_from_env)
```
applecool, line 311: This change is correct right? Because we shouldn't be setting the `ssl_version` value to `self.ssl_version_default_value` at all without having any info on the other `ssl_options`. and always default it to `None` unless there is an env value set. I say this because when none of the certificate attributes are in the env, we set the `ssl_options` to `None` in [line](https://github.com/jupyter/kernel_gateway/blob/19dc96d8fc74568b339248c4c5aad08da4b4c2e5/kernel_gateway/gatewayapp.py#L513). and further more if there are any of the cert attributes mentioned but didn't specify the ssl_version, then we default it to 5 in the else block.

kernel_gateway/tests/test_gatewayapp.py
```diff
@@ -81,7 +81,6 @@ def test_trust_xheaders(self):
 
     def test_ssl_options(self):
         app = KernelGatewayApp()
-        os.environ['KG_CERTFILE'] = '/test/fake.crt'
```
applecool: So, in the test here, we have to have this certfile or key or ca_cert in order to check the default case i.e., if the `ssl_version` is set to `5` or not.

kernel_gateway/gatewayapp.py
kevin-bates: This makes sense.

kernel_gateway/tests/test_gatewayapp.py
kevin-bates: Correct.  I would perform two scenarios in this test.  One that doesn't set the certfile (or keyfile or ca_cert) and asserts ssl_options is None, the other that does set certfile and asserts the default.

applecool: Right :) Cool. I'll make that change :)

applecool:
Thank you :) Should I create the issue on that repo or can I just reference the above issue in that repo?

kevin-bates:
Referencing the issue in a pull request is fine.  Thank you!!

kevin-bates:
Thanks for your contributions @applecool - I'll try to cut 2.4.3 by end of day tomorrow.

applecool:
Awesome. Thank you so much. Really appreciate it :)

kevin-bates:
JKG 2.4.3 is now up on pypi.

applecool:
Hey @kevin-bates  Is it also available on conda `default` or `conda-forge` channel?

kevin-bates:
It should be on conda-forge although I just now merged the PR.

applecool:
Sweet. I can see it on `conda-forge` :) I believe the main channel (default) would be next 😃

PR #346, opened 2020-09-27
Formally define traits in swagger parser

kevin-bates:
The CI tests for Python 3.7 and 3.8 have started failing recently.  It turns out the traitlets 5.0.4 release has removed some deprecated functionality from traitlets 4.2.  Since the older python tests still use 4.3.3, they continue working.  The removed functionality had to do with member fields getting initialized from same-named keyword arguments without having been "typed" as a trait.

By defining these fields as traits, the expected initialization from keyword arguments occurs (on both versions of traitlets).

### DIFF
--- a/kernel_gateway/notebook_http/cell/parser.py
+++ b/kernel_gateway/notebook_http/cell/parser.py
@@ -4,6 +4,7 @@
 
 import re
 import sys
+from traitlets import Unicode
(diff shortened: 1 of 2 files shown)


kevin-bates:
I've gone ahead with an unreviewed merge to try to address doc build issues more clearly.

Issue #352, opened 2021-04-05
local notebook file with URL characters in path break loading


bgerrity:
https://github.com/jupyter/kernel_gateway/blob/7636f219d74309d7888096ef1cedc32eec73655d/kernel_gateway/gatewayapp.py#L352-L375

Special characters can be valid unix filenames; `KernelGatewayApp._load_notebook` incorrectly handles cases where they are used, effectively dropping part of a path.

e.g. `/foo/bar/weird?notebook#name` would be opened as `/foo/bar/weird` despite `weird?notebook#name` being the file's correct name.

kevin-bates:
Fixed via #353

PR #353, opened 2021-04-05
Fix loading of local notebooks with url special characters in path

bgerrity:
Aligns behavior with the comment and fixes the [issue](https://github.com/jupyter/kernel_gateway/issues/352).


kevin-bates (commented):
This change makes sense - thank you.  

Could you please add a test for this?  Seems like just renaming one of the notebooks used as a seed may be sufficient.

kevin-bates (approved):
Thank you for adding the test and this contribution.  I believe the test failures are unrelated.  
(This repo could use some TLC!)

Issue #351, opened 2021-04-01
Add the actual url with an example to the docs?


cottrell:
I've used this before and can't remember what the actual url path is or where to find it. 

A very very specific CURL example should be at the top right after starting it. 

! curl something

from within the notebook.

cottrell:
I think a commnet of look for the port and logging once you launch the process would help people save some time. 

```
[KernelGatewayApp] Registering resource: /hello/world, methods: (['GET'])
[KernelGatewayApp] Registering resource: /_api/spec/swagger.json, methods: (GET)
[KernelGatewayApp] The port 8888 is already in use, trying another port.
[KernelGatewayApp] Jupyter Kernel Gateway at http://127.0.0.1:8889
INFO:tornado.access:200 GET /_api/spec/swagger.json (127.0.0.1) 0.55ms
```

If somebody wants a doc contrib I can just do that.

kevin-bates:
Hi @cottrell - this seems generally useful and a really good idea!  

I think adding a `curl` example from the shell rather than from a notebook would be more applicable since an appeal of the gateway projects is that they're not necessarily accessed by a Notebook/Lab front-end.  The command would be a quick sanity test that the webserver is responsive.

I would recommend using `GET` against the `api/kernelspecs` endpoint as that's non-intrusive and should return something.

Thank you for offering to do this.

kevin-bates:
Fixed via #356

PR #356, opened 2021-04-17
Update docs to be more informative on startup as per issue https://gi…

cottrell:
…thub.com/jupyter/kernel_gateway/issues/351

See https://github.com/jupyter/kernel_gateway/issues/351. 

I built the docs but didn't seem to create a diff so I'm assuming there is some doc build in the github pipelines.

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -3,7 +3,8 @@
 This document describes some of the basics of installing and running the
 Jupyter Kernel Gateway.
 
-## Using pip
+## Install
+### Using pip
 
 We make stable releases of the kernel gateway to PyPI. You can use `pip` to install the latest version along with its dependencies.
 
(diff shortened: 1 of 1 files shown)


kevin-bates (commented):
Hi @cottrell - thank you for the pull request - this is great information.  My only comment is that I wonder if most of this belongs in the `Running` section: https://jupyter-kernel-gateway.readthedocs.io/en/latest/http-mode.html#running

Either that or I think we should make a point that these updates are relative only to the `notebook-http` mode and perhaps add an analogous entry for `jupyter-websocket` mode (which is the default).

Does that sound ok?

cottrell:
Yes, what is the test for the websocket mode? I can add that in.

kevin-bates:
Thank you David.

The startup output is a bit quieter...
```
jupyter kernelgateway --KernelGatewayApp.api=kernel_gateway.jupyter_websocket --port=10100
[KernelGatewayApp] Jupyter Kernel Gateway at http://127.0.0.1:10100
```
(and this option `--KernelGatewayApp.api=kernel_gateway.jupyter_websocket` is the default, but listed for completeness.)

You could hit the same swagger endpoint.  I  also like fetching the set of kernelspecs as a litmus test:
```
curl http://127.0.0.1:10100/api/kernelspecs
```
The output is a bunch of JSON that looks much better if piped through `jq` ( `... |  jq`).

cottrell:
Updated again.

kevin-bates (approved):
Good stuff - thanks David!

Issue #354, opened 2021-04-06
Move CI from Travis to Github Actions


kevin-bates:
Travis will be unavailable in the near future and we should move our CI to Github Actions.

kevin-bates:
@willingc or @parente - I don't have privs to deal with administering Travis for this repo.  Are you able to add that for me?

PR #357, opened 2021-04-18
Replace TravisCI with Github Actions

dolfinus:
TravisCI will soon become dead:
"Please be aware travis-ci.org will be shutting down in several weeks"

Let's move CI to Github Actions. Here is an example of successful workflow: https://github.com/dolfinus/kernel_gateway/actions/runs/761320926

Fixes https://github.com/jupyter/kernel_gateway/issues/354


.github/workflows/tests.yml
```diff
@@ -0,0 +1,82 @@
...
+          do_not_skip: '["pull_request"]'
+
+  tests:
+    name: Run tests (Python ${{matrix.python}})
+    needs: check_duplicate_runs
+    if: ${{ needs.check_duplicate_runs.outputs.should_skip != 'true' }}
+
+    strategy:
+      matrix:
+        python:
```
dolfinus, line 33: I've removed python3.5 from text matrix because `jupyter_client` dependency does not support it:
https://travis-ci.org/github/jupyter/kernel_gateway/jobs/766251855
```
from jupyter_client.channels import HBChannel
  File "/home/travis/virtualenv/python3.5.10/lib/python3.5/site-packages/jupyter_client/channels.py", line 47
    time_to_dead: float = 1.0
                ^
SyntaxError: invalid syntax
```

Also I've added python3.9 to the matrix because it was released recently.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,82 @@
+name: Tests
+
+on:
+  push:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  check_duplicate_runs:
```
dolfinus, line 10: There is an annoying bug of Github Actions - when someone creates a new branch and pushes some commits here, and then creates a pull request, there are 2 copies of workflow running - one for a push event and another for created PR.

This step prevents the second copy of workflow to be run - if tests are already passed for push event then they will be not run twice while PR is created. Also they will not be run if source code hasn't been changed.

kernel_gateway/tests/test_jupyter_websocket.py
```diff
@@ -252,7 +252,7 @@ def test_auth_token(self):
         except Exception as ex:
             self.assertEqual(ex.code, 401)
         else:
-            self.assert_(False, 'no exception raised')
+            raise AssertionError('no exception raised')
```
dolfinus, line 255: `self.assert_` is deprecated, but replacing it with `self.assertTrue(False, ...)` looks ugly. So I've replaces it with just raising an exception.

kernel_gateway/tests/test_jupyter_websocket.py
```diff
@@ -413,13 +413,13 @@ def test_kernel_comm(self):
         }))
 
         # Assert the reply comes back. Test will timeout if this hangs.
-        for _ in range(3):
+        for _ in range(10):
```
dolfinus, line 416: Tests were failed with just 3 retries, so I've increased this value.

kernel_gateway/tests/test_gatewayapp.py
```diff
@@ -92,7 +92,7 @@ def test_ssl_options(self):
         self.assertEqual(ssl_options['ssl_version'], 5)
 
     def test_load_notebook_local(self):
-        nb_path = os.path.join(RESOURCES, 'weirdly?named#notebook.ipynb')
```
dolfinus, line 95: This file name causes an error while cloning the repo on a Windows machine - NTFS does not allow to use `?` in file names.
So I've renamed it.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,82 @@
...
+    - name: Show help
+      run: jupyter kernelgateway --help
+
+    - name: Run tests
+      run: nosetests --process-restartworker --with-coverage --cover-package=kernel_gateway
+      env:
+        ASYNC_TEST_TIMEOUT: 10
+
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v1
```
dolfinus, line 80: This new step is present in jupyter_client CI

blink1073 (approved):
Amazing work, thank you!

kevin-bates:
> Amazing work, thank you!

Yes, agreed - thank you @dolfinus!  (Thanks @blink1073 for reviewing this.)

Issue #297, opened 2018-07-26
500 Error when Changing Kernel through Session


bashterm:
The kernel gateway returns a 500 error when I try to change the Kernel in the current session.

I'm running the kernel gateway on docker.

Here is the server log:
```
jupyter_1   | [KernelGatewayApp] Starting buffering for c0169618-1f04-49fe-8b98-311f6df27668:270b28ea405236eaf399c29bf4be76e6
jupyter_1   | [E 180726 19:12:54 web:1621] Uncaught exception OPTIONS /api/sessions/df4015c9-9824-4c6f-854b-fe8601f0b148?1532632374048 (172.18.0.1)
jupyter_1   |     HTTPServerRequest(protocol='http', host='localhost:8888', method='OPTIONS', uri='/api/sessions/df4015c9-9824-4c6f-854b-fe8601f0b148?1532632374048', version='HTTP/1.1', remote_ip='172.18.0.1')
jupyter_1   |     Traceback (most recent call last):
jupyter_1   |       File "/opt/conda/lib/python3.6/site-packages/tornado/web.py", line 1541, in _execute
jupyter_1   |         result = method(*self.path_args, **self.path_kwargs)
jupyter_1   |     TypeError: options() got an unexpected keyword argument 'session_id'
jupyter_1   | [KernelGatewayApp] WARNING | Unhandled error
jupyter_1   | [E 180726 19:12:54 web:2106] 500 OPTIONS /api/sessions/df4015c9-9824-4c6f-854b-fe8601f0b148?1532632374048 (172.18.0.1) 2.04ms
```

Any help or advice would be appreciated.

kevin-bates:
I believe more information is required here.  What does your request look like?  I'm curious where the id `df4015c9-9824-4c6f-854b-fe8601f0b148?1532632374048` is coming from since just prior to that, we see `c0169618-1f04-49fe-8b98-311f6df27668:270b28ea405236eaf399c29bf4be76e6`, the first portion of which presumably refers to the original kernel_id.

When I try changing kernels from within a Notebook configured with NB2KG to Enterprise Gateway (which should be the same behavior), I can successfully do that.  However, the REST calls I see are all against `/api/kernels`, not `/api/sessions`.  In addition, the sequence is to delete the original kernel, then create a new kernel (with a new id) via `DELETE` and `POST` calls respectively.

[...]

PR #363, opened 2021-05-24
Add preflight and delete option in sessions handler

damiles:
When try to call a delete session, /api/sessions/<session_id> we obtain a preflight error and cors. See screenshots for more info:

image

image

I solve adding a new handler and allowing options call for preflight and adding delete options in headers. (Maybe is not all correct for integration because I don't read full kernel_gateway code) But this works fine and solve the issue.

This issue looks like appears in #297 too.

### DIFF
--- a/kernel_gateway/services/sessions/handlers.py
+++ b/kernel_gateway/services/sessions/handlers.py
@@ -27,6 +27,23 @@ class SessionRootHandler(TokenAuthorizationMixin,
         else:
             super(SessionRootHandler, self).get()
 
+class SessionHandler(TokenAuthorizationMixin,
+                    CORSMixin,
+                    JSONErrorsMixin,
+                    notebook_handlers.SessionHandler):
+    """Extends the notebook session handler with token auth, CORS, and
+    JSON errors.
+    """
+    def set_default_headers(self):
+        self.set_header('Content-Type', 'application/json')
+        self.set_header('Access-Control-Allow-Origin', '*')
+        self.set_header('Access-Control-Allow-Headers', 'content-type')
+        self.set_header('Access-Control-Allow-Methods', 'DELETE')
+
+    def options(self, **kwargs):
+        """Method for properly handling CORS pre-flight"""
+        self.finish()
+
 default_handlers = []
 for path, cls in notebook_handlers.default_handlers:
     if cls.__name__ in globals():


kevin-bates:
@blink1073 or @lresende, I'm hoping one of you could take a look at this PR.  I just don't know enough about this aspect of the stack to know if this introduces any issues or not - sorry about that.  (Thank you)

lresende (approved):

lresende:
I haven't actually tested but did a little research on pre-flight calls and the code looks good.

kernel_gateway/services/sessions/handlers.py
```diff
@@ -27,6 +27,23 @@ def get(self):
...
+class SessionHandler(TokenAuthorizationMixin,
+                    CORSMixin,
+                    JSONErrorsMixin,
+                    notebook_handlers.SessionHandler):
+    """Extends the notebook session handler with token auth, CORS, and
+    JSON errors.
+    """
+    def set_default_headers(self):
+        self.set_header('Content-Type', 'application/json')
+        self.set_header('Access-Control-Allow-Origin', '*')
```
blink1073, line 39: I think this needs to be set based on the `allow_origin` trait: https://github.com/jupyter-server/kernel_gateway/blob/5269af0a2bf84b8a93b0c983d731b82843daec02/kernel_gateway/gatewayapp.py#L164

kevin-bates: @blink1073 - so something [like this](https://github.com/jupyter-server/kernel_gateway/blob/5269af0a2bf84b8a93b0c983d731b82843daec02/kernel_gateway/gatewayapp.py#LL484C9-L484C9)?
```suggestion
        self.set_header('Access-Control-Allow-Origin', self.settings['kg_allow_origin'])
```

blink1073: Yep, that should do it.

kevin-bates: @blink1073  - Hmm, sorry, but should [allow-methods and allow-headers be handled similarly](https://github.com/jupyter-server/kernel_gateway/blob/5269af0a2bf84b8a93b0c983d731b82843daec02/kernel_gateway/gatewayapp.py#L482-L483) - since these are configurable as well?

It's really not clear to me what this is all about.

blink1073: Yeah, we should use those as well.

kevin-bates: Thanks - those changes have been pushed.

blink1073 (approved):

kevin-bates:
Hi @damiles - I know it's been "forever" since you opened this PR and we apologize for not being more active on this repository.  I'd like to get this PR merged (and released via a patch release `2.5.3`) so we can look into creating a 3.0 release that uses Jupyter Server rather than notebook for its foundation.

Are you able to try out the changes we've been discussing that I just pushed?  Since the notebook/jupyter server gateway client doesn't use the Session API when interacting with the Gateway server, I suspect you must have your own frontend that this PR is addressing.

damiles:
@kevin-bates Don't worry, I will test it on my own frontend as soon as possible.

kevin-bates:
Awesome - thank you @damiles!  Please let us know how it goes and we'll cut a release as soon as this PR is merged.

kernel_gateway/services/sessions/handlers.py
```diff
@@ -27,6 +27,23 @@ def get(self):
...
+    """
+    def set_default_headers(self):
+        self.set_header('Content-Type', 'application/json')
+        self.set_header('Access-Control-Allow-Origin', self.settings['kg_allow_origin'])
+        self.set_header('Access-Control-Allow-Headers', self.settings['kg_allow_headers'])
+        self.set_header('Access-Control-Allow-Methods', self.settings['kg_allow_methods'])
+
+    def options(self, **kwargs):
+        """Method for properly handling CORS pre-flight"""
+        self.finish()
```
kevin-bates, line 45: After working on #384, I'm curious why the [existing `CORSMixin`](https://github.com/jupyter-server/kernel_gateway/blob/7d592e737bf17dddd4464eab9c2f3cd47dd04507/kernel_gateway/mixins.py#L14-L45) isn't already sufficient, since it adds everything that is listed here, except `Content-Type`.

Are the `bases` of the `else` statement below not doing what is intended?

PR #368, opened 2021-08-21
use get_shell_msg helper method

### LINKED ISSUE #367: block kwarg removed in jupyter-client 7
[This commit](https://github.com/jupyter/jupyter_client/commit/f365a22a22591c3bae04723b804585fe11934ce5) in jupyter-client 7 removes the `block` kwarg for `get_msg` used [here](https://github.com/jupyter/kernel_gateway/blob/213a8179c5a6e81a301cafc467549ef90855468f/kernel_gateway/services/kernels/manager.py#L104)

tschaume:
fixes #367

### DIFF
--- a/kernel_gateway/services/kernels/manager.py
+++ b/kernel_gateway/services/kernels/manager.py
@@ -101,7 +101,7 @@ class SeedingMappingKernelManager(MappingKernelManager):
                     # executed
                     if self.parent.personality.should_seed_cell(code):
                         client.execute(code)
-                        msg = client.shell_channel.get_msg(block=True)
+                        msg = client.get_shell_msg(block=True)
                         if msg['content']['status'] != 'ok':
                             # Shutdown the channels to remove any lingering ZMQ messages
                             client.stop_channels()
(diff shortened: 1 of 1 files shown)


kernel_gateway/services/kernels/manager.py
```diff
@@ -101,7 +101,7 @@ def start_kernel(self, *args, **kwargs):
                     # executed
                     if self.parent.personality.should_seed_cell(code):
                         client.execute(code)
-                        msg = client.shell_channel.get_msg(block=True)
+                        msg = client.get_shell_msg(block=True)
```
kevin-bates, line 104: I think the `block=True` needs to be removed.

@davidbrochart - Should we update the [migration guide](https://jupyter-client.readthedocs.io/en/latest/migration.html#zmqsocketchannel) to indicate what calls using `block=True` should do since we only address the `block=False` case wrt to a `timeout` value?  I found removing the `block=True` to be successful but am not entirely sure.

davidbrochart: `block=True` was the default value, so it could have been omitted previously. It translates to a value of `timeout>0` now, which is the default value, so there's nothing to do, and `block=True` just has to be removed. We can clarify this in the migration guide, and also state that using e.g. `client.get_shell_msg()` should be used instead of `client.shell_channel.get_msg()`. I opened https://github.com/jupyter/jupyter_client/pull/685.

kevin-bates (approved):

kevin-bates:
Hi @tschaume - I went ahead and pushed the change to remove the `block` parameter and the tests now pass.  

Once this is merged, I'd like to cut 2.5.1.

PR #375, opened 2022-12-13
correct create_personality function params to reflect usage

bgerrity:
The names of the positions args used in the docs don't reflect [usage](https://github.com/jupyter-server/kernel_gateway/blob/c72803ab1ea4bb75847917e0a7c682acbe700fbe/kernel_gateway/gatewayapp.py#L447); this leads to a `TypeError`. This change address that issue.

```
  File ".../gateway#link-tree/kernel_gateway/gatewayapp.py", line 447, in init_configurables
    self.personality = func(parent=self, log=self.log)
TypeError: create_personality() got an unexpected keyword argument 'log'
````

### DIFF
--- a/docs/source/plug-in.md
+++ b/docs/source/plug-in.md
@@ -33,7 +33,7 @@ class TemplatePersonality(LoggingConfigurable):
 
 def create_personality(parent, log):
     """Put docstring here."""
-    return TemplatePersonality(parent=parent)
+    return TemplatePersonality(parent=parent, log=log)
 ```
 
 Provided personalities include [kernel_gateway.jupyter_websocket](websocket-mode.md) and [kernel_gateway.notebook_http](http-mode.md).
--- a/docs/source/plug-in.md
+++ b/docs/source/plug-in.md
@@ -31,7 +31,7 @@ class TemplatePersonality(LoggingConfigurable):
         been specified."""
         pass
 
-def create_personality(self, parent):
+def create_personality(parent, log):
     """Put docstring here."""
     return TemplatePersonality(parent=parent)
 ```
--- a/docs/source/plug-in.md
+++ b/docs/source/plug-in.md
@@ -31,9 +31,9 @@ class TemplatePersonality(LoggingConfigurable):
         been specified."""
         pass
 
-def create_personality(parent, log):
+def create_personality(*args, **kwargs):
     """Put docstring here."""
-    return TemplatePersonality(parent=parent, log=log)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by bgerrity** [comment]
Broken tests appear unrelated. Local build of docs succeeds.

File: docs/source/plug-in.md [hunk 1]
```diff
@@ -31,7 +31,7 @@ class TemplatePersonality(LoggingConfigurable):
         been specified."""
         pass
 
-def create_personality(self, parent):
+def create_personality(parent, log):
```
kevin-bates, line 34: Good catch!  

Should the signature be: `def create_personality(*args, **kwargs):` and we document two required kwargs are: `parent` and `log`?  This will then fit the existing declarations that reside for the [websocket](https://github.com/jupyter-server/kernel_gateway/blob/c72803ab1ea4bb75847917e0a7c682acbe700fbe/kernel_gateway/jupyter_websocket/__init__.py#L81) and [http](https://github.com/jupyter-server/kernel_gateway/blob/c72803ab1ea4bb75847917e0a7c682acbe700fbe/kernel_gateway/notebook_http/__init__.py#L156) personalities while allowing for other parameters to be added in the future.

bgerrity: I'm unclear about the usage of `*args, **kwargs` here: Based on [`init_configurables`](https://github.com/jupyter-server/kernel_gateway/blob/c72803ab1ea4bb75847917e0a7c682acbe700fbe/kernel_gateway/gatewayapp.py#L447), I'm not I'm not seeing how any additional args could be passed to `create_personality` via configuration.

kevin-bates: In the future, we may find the need to convey other information to a "personality", at which time we'd add a parameter (e.g., `foo=bar`) in `init_configurables`.  Personalities that don't know about the parameter could still be called (and ignore it), while those that know what `foo` is for would use it via `kwargs["foo"]`, for example.

This, coupled with the idea that signatures for "custom personalities" should match that of the two existing personalities, leads me to argue for using `*args, **kwargs` due to precedent.

One of the things that makes this confusing is the lack of initializers with "Configurable" classes - on which most of Jupyter is based.  Both `parent` and `log` are "well-known" parameters within the `LoggingConfigurable` class hierarchy.  (Not sure if this helps or not.)

bgerrity: Makes sense to use that for convention and extensibility.

kevin-bates (approved):
LGTM - sorry about the CI.  This project needs substantial maintenance!

PR #406, opened 2024-12-13
Add support for displaying image/png data.

teeler:
This change maps image/png display data from execution results as  tags with in-line data.

Side note - Hi! I'm Tyler, I work on Colab at Google. Huge fans of this project - we're using it to show folks how to deploy Colab notebooks on Cloud Run with the kernel gateway (https://x.com/GoogleColab/status/1864765949913739725).

I added this bit of functionality so users could display simple (and only png) objects in notebooks, for example this pair of cells:

```
# ResponseInfo GET /plot
print(
    json.dumps(
        {
            "headers": {"Content-Type": "text/html"},
            "status": 200,
        }
    )
)
```

And

```
# GET /plot

# prompt: Generate a random array with numpy and plot it (thanks Gemini!)

import numpy as np
import matplotlib.pyplot as plt

# Generate a random array
random_array = np.random.rand(10)

# Plot the array
plt.plot(random_array)
plt.title("Random Array Plot")
plt.xlabel("Index")
plt.ylabel("Value")
plt.show()
```

Produces correct output for /plot 

Screenshot 2024-12-13 at 3 17 31 PM

as an in-line `` tag. 

```
<img alt="<Figure size 640x480 with 1 Axes>" src="data:image/png;base64,iVBORw....." >
```

It also has the capability of mixing between images/html/text. 

Happy to re-arrange/format/etc. any of this code, thanks for reading!

### DIFF
--- a/kernel_gateway/notebook_http/handlers.py
+++ b/kernel_gateway/notebook_http/handlers.py
@@ -61,6 +61,32 @@ class NotebookAPIHandler(
         self.response_sources = response_sources
         self.kernel_language = kernel_language
 
+    def _accumulate_display(self, results):
+        """Accumulates result chunks for "display" messages and prepares them as
+
+        in-line <img> tags. See
+        https://ipython.org/ipython-doc/3/development/messaging.html#display-data
+        for details on the display protocol.
+
+        Parameters
+        ----------
+        results: list
+          A list of results containing display data.
+        """
+        out = []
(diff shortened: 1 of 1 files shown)


teeler:
@parente (sorry for tagging you directly) - I'd be happy to fix up some of these tests in exchange for some feedback on this PR. ;)

